PR #11, opened 2020-09-23
Process pipeline richard

richardyz98:
Here I finished testing for collapseUMI (both sorting and collapsing) and collapseBam2DF using a bam file that I generated from hand, testing multiple possible cases. I also reformatted all of the files using black, with line length 80, and with 4 spaces as indent.


mattjones315 (approved):
Looks great to me!

PR #15, opened 2020-10-01
Process pipeline richard

richardyz98:
Made some formatting changes to the error correct UMI related files, including naming conventions and logging


mattjones315 (changes requested):
Table the tie breaking in error correct umis for now.

mattjones315 (approved):
Looks good now, thanks!

PR #16, opened 2020-10-02
Call Alleles

mattjones315:
Converted perl script for calling indels from cigar strings into python 
Wrote tests
Wrapped the cigar parsing function into a larger function that processes a series of alignments (might be nice to parallelize in the future, but the SW aligner is already quite fast)


cassiopeia/ProcessingPipeline/process/pipeline.py
```diff
@@ -301,10 +303,82 @@ def align_sequences(
...
+    for _, row in tqdm(
+        alignments.iterrows(),
+        total=alignments.shape[0],
+        desc="Parsing CIGAR strings into indels",
+    ):
+
+        alignment_to_indel[row.readName] = alignment_utilities.parse_cigar(
+            row.CIGAR,
+            row.Seq,
+            ref,
```
richardyz98, line 369: Hi Matt I think this is missing the start_pos argument

cassiopeia/ProcessingPipeline/process/alignment_utilities.py
```diff
@@ -0,0 +1,214 @@
...
+    indels = ["" for site in cutsites]
+    uncut_indels = ["" for site in cutsites]
+    intBC = ""
+
+    ref_anchor = ref[barcode_interval[0] - 11 : barcode_interval[0]]
+    intBC_length = barcode_interval[1] - barcode_interval[0]
+
+    # set up two pointers to the reference and query, respectively
+    ref_pointer = start_pos
+    query_pointer = 0
```
richardyz98: I think we should initialize query_pointer to query_begin found in the documentation for skbio's AlignmentStructure:
http://scikit-bio.org/docs/0.5.1/generated/skbio.alignment.AlignmentStructure.html#skbio.alignment.AlignmentStructure
and also the ref_pointer to target_begin. These can be passed in from call_indels

richardyz98 (changes requested):

richardyz98 (approved):

PR #24, opened 2020-10-16
CassiopeiaSolver & GreedySolver Skeleton

mattjones315:
Implements abstract class CassiopeiaSolver
Skeleton code for GreedySolver, a subclass of CassiopeiaSolver

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,2 +1,16 @@
+version: 2
+build:
+  image: latest
+sphinx:
+  configuration: docs/conf.py
 python:
-    version: 3.6
+  version: 3.6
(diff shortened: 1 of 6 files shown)


richardyz98 (approved):
Looks good Matt! Actually though, I think that some of the top-down methods don't necessarily split on one character, so I think that we don't need perform split. find_split would then return simply the subset of samples in the left side of the split, and the subset of the samples in the right side of the split, and would handle missing data. Then the tree building process would be as simple as appending the samples on each side to an internal node for each split (with maybe some handling of duplicates). I can change up the code in GreedySolver. What do you think?

PR #27, opened 2020-10-31
Processing pipeline

mattjones315:
Implemented an abstract class `DistanceSolver` and a subclass `NeighborJoiningSolver`.

### DIFF
--- a/cassiopeia/solver/DistanceSolver.py
+++ b/cassiopeia/solver/DistanceSolver.py
@@ -1,0 +1,147 @@
+"""
+This file stores a subclass of CassiopeiaSolver, the DistanceSolver. Generally,
+the inference procedures that inherit from this method will need to implement
+methods for selecting "cherries" and updating the dissimilarity map. Methods
+that will inherit from this class by default are Neighbor-Joining and UPGMA.
+There may be other subclasses of this
+"""
+import abc
+import networkx as nx
+import numpy as np
+import pandas as pd
+from typing import Callable, Dict, Optional, Tuple
+
+from cassiopeia.solver import CassiopeiaSolver
+
+
+class DistanceSolveError(Exception):
+    """An Exception class for all DistanceSolver subclasses.
+    """
+
+    pass
+
+
+class DistanceSolver(CassiopeiaSolver.CassiopeiaSolver):
+    def __init__(
+        self,
+        character_matrix: pd.DataFrame,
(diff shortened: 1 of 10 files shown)


cassiopeia/solver/DistanceSolver.py
```diff
@@ -0,0 +1,147 @@
...
+            )
+
+            N = _dissimilarity_map.shape[0]
+
+        new_node_name = len(tree.nodes)
+        tree.add_node(new_node_name)
+
+        tree.add_edges_from([(new_node_name, i) for i in _dissimilarity_map.index])
+
+        tree = nx.relabel_nodes(tree, identifier_to_sample)
```
richardyz98, line 111: Is there a reason to use the named indices of the cm and then relabel them to the numerical indices instead of just directly using the numerical indices?

richardyz98, line 111: Is there a reason to use the named indices of the cm and then relabel them to the numerical indices instead of just directly using the numerical indices?

richardyz98 (changes requested):

cassiopeia/solver/DistanceSolver.py
mattjones315: I think this is no longer an issue in the current code - I actually just add all the samples by name directly to the graph. If this isn't scalable enough, I'll only add unique samples and add a post-processing routine afterwards.

cassiopeia/solver/NeighborJoiningSolver.py
```diff
@@ -14,15 +15,66 @@
...
         dissimilarity_map: Optional[pd.DataFrame] = None,
         dissimilarity_function: Optional[Callable] = None,
+        root_sample: Optional[str] = None,
     ):
 
+        if dissimilarity_function is None and root_sample is None:
+            raise DistanceSolver.DistanceSolveError(
+                "Please specify a root sample or provide a dissimilarity "
+                "function by which to add a root to the dissimilarity map"
+            )
```
richardyz98, line 68: Is this check necessary? It doesn't seem you're using the dissimilarity function to impute a root, and if the root_sample is None then you add the root.

richardyz98 (changes requested):

cassiopeia/solver/NeighborJoiningSolver.py
mattjones315: Fixed in the newest commit!

PR #36, opened 2020-11-20
Process pipeline richard

richardyz98:
Added maxcutsolver and graph_utils, as well as tests for the maxcut


cassiopeia/solver/graph_utilities.py
```diff
@@ -0,0 +1,361 @@
...
+        missing_char: The character representing missing values
+        samples: A list of samples to build the graph over
+        w: A set of optional weights for edges in the connectivity graph
+
+    Returns:
+        A connectivity graph constructed over the sample set
+    """
+    G = nx.Graph()
+    if not samples:
+        samples = range(cm.shape[0])
```
mattjones315, line 59: Realistically, do we ever invoke this procedure without passing samples?

cassiopeia/solver/graph_utilities.py
```diff
@@ -0,0 +1,361 @@
...
+        )
+    else:
+        return sum(
+            [
+                w[i][cm.iloc[u, i]]
+                for i in range(k)
+                if cm.iloc[u, i] == cm.iloc[v, i]
+                and (cm.iloc[u, i] != "0" and cm.iloc[u, i] != missing_char)
+            ]
+        )
```
mattjones315, line 361: Let's take this out for now - only keep the code that's relevant for the MaxCut algorithm PR. This is hard to review without knowing how it's used in the spectral method.

test/solver_tests/maxcut_test.py
```diff
@@ -0,0 +1,63 @@
+import unittest
+
+import networkx as nx
+import pandas as pd
+
+from cassiopeia.solver.MaxCutSolver import MaxCutSolver
+from cassiopeia.solver import graph_utilities
+from cassiopeia.solver import solver_utilities
+
+class MaxCutSolverTest(unittest.TestCase):
```
mattjones315, line 10: You should add some more tests here around the helper function `check_if_cut`. Also I know that this algorithm is stochastic, but there must be a very simple case that it's guaranteed to find the right tree with (e.g. PP?). You should write a test making sure it passes this simple case.

test/solver_tests/maxcut_test.py
```diff
@@ -0,0 +1,63 @@
+import unittest
+
+import networkx as nx
+import pandas as pd
+
+from cassiopeia.solver.MaxCutSolver import MaxCutSolver
+from cassiopeia.solver import graph_utilities
+from cassiopeia.solver import solver_utilities
```
mattjones315, line 8: `solver_utilities` doesn't seemed to be used here. Remove please.

mattjones315 (changes requested):
Looks good! A few comments on documentation (though it generally looks fantastic!) and code readability. You should add some more tests around the helper functions for MaxCut and perhaps a _very_ simple base case. Also please remove the code in `graph_utilities.py` that's not being used in this PR.

cassiopeia/solver/MaxCutSolver.py
```diff
@@ -0,0 +1,169 @@
...
+            meta_data: Data table storing meta data for each sample
+            priors: Prior probabilities of character state transitions
+            tree: The tree built by `self.solve()`. None if `solve` has not been
+                called yet
+            prune_cm: A character matrix with duplicate rows filtered out, removing
+                doublets from the sample set
+            sdimension: The number of dimensions to use for the embedding space
+            iterations: The number of iterations in updating the embeddings
+            weights: A set of optional weights for edges in the connectivity graph
+        """
```
mattjones315, line 65: Still not in the right place. Put this _between_ the class declaration and the init function declaration. see my code in the NeighborJoiningSolver for an example.

mattjones315 (commented):
All looks good! Just make that one documentation adjustment and then w're good to go.

mattjones315 (approved):
Looks great!

PR #37, opened 2020-11-20
added critique library and tests

mattjones315:


cassiopeia/critique/compare.py
```diff
@@ -0,0 +1,127 @@
...
+
+        all_triplets_correct[depth] /= number_of_trials
+        resolvable_triplets_correct[depth] /= (
+            number_of_trials - number_unresolvable_triplets
+        )
+        
+        if number_unresolvable_triplets == 0:
+            unresolved_triplets_correct[depth] = 1.0
+        else:
+            unresolved_triplets_correct[depth] /= max(1, number_unresolvable_triplets)
```
richardyz98, line 116: Is the max necessary?

cassiopeia/critique/compare.py
```diff
@@ -0,0 +1,127 @@
...
+
+            all_triplets_correct[depth] += score
+            if is_resolvable:
+                resolvable_triplets_correct[depth] += score
+            else:
+                unresolved_triplets_correct[depth] += score
+
+        all_triplets_correct[depth] /= number_of_trials
+        resolvable_triplets_correct[depth] /= (
+            number_of_trials - number_unresolvable_triplets
```
richardyz98, line 110: Maybe we should also create a case where all of the triplets are unresolvable? In which case we would want to do the 0 denominator check as well

cassiopeia/critique/compare.py
mattjones315: No, good catch! I'll remove it.

cassiopeia/solver/solver_utilities.py
```diff
@@ -165,15 +166,42 @@ def collapse_tree(
...
+
+    collapse_fn = lambda x: (len(x.children) == 1)
+
+    collapsed_tree = tree.copy()
+    to_collapse = [n for n in collapsed_tree.traverse() if collapse_fn(n)]
+
+    for n in to_collapse:
+        n.delete()
+
+    return collapsed_tree
```
richardyz98, line 190: Are the children of the nodes attached to the parent of the deleted node?

test/solver_tests/edge_collapse_test.py
```diff
@@ -137,6 +138,42 @@ def test_newick_converter(self):
...
+        collapsed_tree = solver_utilities.collapse_unifurcations(tree)
+
+        # make sure all leaves remain
+        self.assertEqual(len(tree), len(collapsed_tree))
+        for n in tree:
+            self.assertIn(n.name, collapsed_tree.get_leaf_names())
+
+        # make sure there are no singletons left
+        for n in collapsed_tree.traverse():
+            self.assertFalse(len(n.children) == 1)
```
richardyz98, line 157: Maybe we should also test that the final tree is connected, and that the nodes that should be connected are connected?

richardyz98 (changes requested):

test/solver_tests/edge_collapse_test.py
richardyz98: for example making sure that node 0 is connected to node 3 in the final tree?

cassiopeia/solver/solver_utilities.py
mattjones315: Yes, this is done whenever you delete a given node: http://etetoolkit.org/docs/latest/reference/reference_tree.html#ete3.TreeNode.delete

test/solver_tests/edge_collapse_test.py
mattjones315: Done.

cassiopeia/critique/compare.py
mattjones315: Done.

richardyz98 (approved):

PR #40, opened 2020-11-25
Process pipeline richard

richardyz98:
Everything Spectral


cassiopeia/solver/graph_utilities.py
```diff
@@ -160,3 +160,213 @@ def max_cut_improve_cut(G: nx.DiGraph, cut: List[int]):
...
+        )
+    else:
+        return sum(
+            [
+                w[i][cm.iloc[u, i]]
+                for i in range(k)
+                if cm.iloc[u, i] == cm.iloc[v, i]
+                and (cm.iloc[u, i] != "0" and cm.iloc[u, i] != missing_char)
+            ]
+        )
```
mattjones315, line 372: I think we can extend this in two ways:

Firstly, this should be the modified hamming similarity: if two characters are the same, give a +2; if two characters are different and one is zero, give a +1; if two characters are different, give +0. 

Secondly, we can improve this even more using state priors. We can add log-probs of these states so that if two characters are the same we would give + (log(a) + log(b)), etc.

mattjones315 (changes requested):
Looks awesome Richard! I have some questions and comments that I think should be addressed before merging the code in.

cassiopeia/solver/SpectralSolver.py
```diff
@@ -0,0 +1,165 @@
...
+
+    Attributes:
+        character_matrix: The character matrix describing the samples
+        missing_char: The character representing missing values
+        meta_data: Data table storing meta data for each sample
+        priors: Prior probabilities of character state transitions
+        tree: The tree built by `self.solve()`. None if `solve` has not been
+            called yet
+        prune_cm: A character matrix with duplicate rows filtered out, removing
+            doublets from the sample set
```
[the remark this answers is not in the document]
richardyz98: Thanks Matt this is a good point to understand

cassiopeia/solver/SpectralSolver.py
```diff
@@ -0,0 +1,165 @@
...
+        G = graph_utilities.construct_similarity_graph(
+            self.prune_cm,
+            mutation_frequencies,
+            self.missing_char,
+            samples,
+            threshold=self.threshold,
+            w=self.weights,
+        )
+
+        L = nx.normalized_laplacian_matrix(G).todense()
```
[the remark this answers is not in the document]
richardyz98: I think this is required to be dense for sp.linalg.eig to work

cassiopeia/solver/graph_utilities.py
```diff
@@ -194,8 +195,6 @@ def construct_similarity_graph(
         A similarity graph constructed over the sample set
     """
     G = nx.Graph()
-    if not samples:
-        samples = range(cm.shape[0])
     for i in samples:
```
mattjones315, line 198: In the function definition, you have samples as possibly being an empty list. Let's make this a mandatory argument. But if you don't want to, then technically the typing should be wrapped in `Optional`

cassiopeia/solver/graph_utilities.py
```diff
@@ -238,12 +237,36 @@ def spectral_improve_cut(G: nx.Graph, cut: List[int]) -> List[int]:
         A new partition that is a local minimum to the objective function
     """
 
-    delta_num = {}
-    delta_denom = {}
+    def set_improvement_potential(u):
```
mattjones315, line 240: What is `u`? Add full documentation here.

mattjones315 (commented):
Looking great! Please resolve my two nitpicky comments.

cassiopeia/solver/graph_utilities.py
```diff
@@ -237,26 +237,26 @@ def spectral_improve_cut(G: nx.Graph, cut: List[int]) -> List[int]:
         A new partition that is a local minimum to the objective function
     """
 
-    def set_improvement_potential(u):
+    def set_improvement_potential(node: int):
```
mattjones315, line 240: Please add full documentation to this helper function. This is still missing an Args section, a Returns section, a return type in the function declaration

mattjones315 (commented):
Can you please resolve that last comment?

mattjones315 (approved):
Looks good, thanks1

PR #45, opened 2020-12-11
added weights to all greedy variants, including testing

richardyz98:
Hi Matt,
This PR includes some variable name cleanup, as well as the use of weights for vanilla, maxcut, spectral greedy. Also, I did some testing for weights for all algorithms where weights are applicable


cassiopeia/solver/MaxCutGreedySolver.py
```diff
@@ -92,33 +90,56 @@ def perform_split(
...
+                                mutation_frequencies[character][state]
+                                > best_frequency
+                            ):
+                                chosen_character, chosen_state = (
+                                    character,
+                                    state,
+                                )
+                                best_frequency = mutation_frequencies[
+                                    character
+                                ][state]
```
mattjones315, line 130: This is incorrect. We multiply by the _log_ of these priors. Please refer back to the original Cassiopeia GreedySolver for the correct math.

If I understand this, you'd like to weight the mutation frequency by the prior. In this case, you might want to do something like `-nlog(p)` like we did in the original paper.

cassiopeia/solver/SpectralGreedySolver.py
```diff
@@ -94,33 +92,56 @@ def perform_split(
...
+                                > best_frequency
+                            ):
+                                chosen_character, chosen_state = (
+                                    character,
+                                    state,
+                                )
+                                best_frequency = mutation_frequencies[
+                                    character
+                                ][state]
+
```
mattjones315, line 133: Same as above.

cassiopeia/solver/VanillaGreedySolver.py
```diff
@@ -78,33 +78,56 @@ def perform_split(
...
+                                mutation_frequencies[character][state]
+                                > best_frequency
+                            ):
+                                chosen_character, chosen_state = (
+                                    character,
+                                    state,
+                                )
+                                best_frequency = mutation_frequencies[
+                                    character
+                                ][state]
```
mattjones315, line 118: I think this definitely needs to be changed. Should be `-nlog(p)` as we had in the original paper.

mattjones315 (changes requested):
Looking good! I had some comments you should resolve.

cassiopeia/solver/MaxCutGreedySolver.py
richardyz98: Hi Matt,
So I think this is a discrepancy in the way that we think about these weights.

I think that potentially the fact that I have labeled the weights as priors here is confusing. But the way that I see it, the algorithms should be able to take in generic weights by which to weight the frequencies. These weights can be derived from priors, in which case a dictionary of -np.log(p) weights could be used, or some other transformation on p. But I can also imagine using some other types of weights in the future. What do you think? This would require some preprocessing to create the dictionary of weights, an example of which you can see in the vanillagreedy_test.py testing file. What do you think? Is this reasonable?

mattjones315: Hey Richard,

These are fair points, but we've sort of developed our inference framework in the context of priors being non-logged. As you know, the original Cassiopeia ms & codebase logged these priors when incorporating them into the Greedy solver. For any user using Cassiopeia right now, that's how they're accustomed to using the priors. Plus, in the Cassiopeia codebase we have functionality for calculating the priors as frequencies and then normalizing the probabilities when creating the character matrices. The most straightforward solution here, in my opinion, would be to have user specify the priors as usual and we do the log transformation on the backend as usual.  

An alternative would be to have a `priors` and `weights`. If only priors are specified, we set `self.weights` to be the log of the priors. Else, we take whatever the user passes as weights.

Thoughts?

cassiopeia/solver/GreedySolver.py
```diff
@@ -46,18 +51,26 @@ def __init__(
...
+        priors: Optional[Dict[int, Dict[int, float]]] = None,
+        weights: Optional[Dict[int, Dict[int, float]]] = None,
     ):
 
         super().__init__(character_matrix, missing_char, meta_data, priors)
+        self.weights = weights
+        if priors:
+            self.priors = priors
+            if not weights:
+                self.weights = solver_utilities.negative_log_prob_weights(
```
mattjones315, line 63: Is this a good design decision? Do you imagine doing other fancy things than just applying `np.log10` to all priors?

cassiopeia/solver/dissimilarity_functions.py
```diff
@@ -70,3 +69,49 @@ def weighted_hamming_distance(
...
+            [
+                w[i][s1[i]]
+                for i in range(num_chars)
+                if s1[i] == s2[i]
+                and (
+                    s1[i] != 0
+                    and s1[i] != missing_char
+                )
+            ]
+        )
```
mattjones315, line 117: I think this is going to be redundant with my `hamming_distance` function. Why not use the weighted hamming dissimilarity function that I wrote previously?

cassiopeia/solver/solver_utilities.py
```diff
@@ -228,3 +228,25 @@ def _to_newick_str(g, node):
...
+    Returns:
+        A dictionary of weights for each character/state pair
+    """
+    weights = {}
+    for character in priors:
+        state_weights = {}
+        for state in priors[character]:
+            state_weights[state] = -np.log(priors[character][state])
+        weights[character] = state_weights
+    return weights
```
mattjones315, line 252: Do you think it would be worth generalizing this to something like `transform_priors` that takes in a function to apply to all priors?

mattjones315 (commented):
Hey Richard, can you resolve my comments when you get a chance?

cassiopeia/solver/GreedySolver.py
richardyz98: Yeah Robert and I have floated a couple of ideas for other ways to utilize priors, like taking a 1/(m * q{j}) weight for use in similarity. I think it would be helpful to have this going down the line

cassiopeia/solver/dissimilarity_functions.py
richardyz98: It's slightly different from your weighted hamming dissimilarity in that it only increments the score +1 when you have a match, when in the case of the weighted hamming it increments +2. And it doesn't have the case where if the characters are different but one of them is 0, then it increments +1. I think that this will likely not be the best similarity to use for spectral in the future, but I kept it for now. When we get benchmarks that show that the weighted hamming distance allows for better performance, I think then I'll remove it

richardyz98: It's a more naive notion of similarity for sure

cassiopeia/solver/solver_utilities.py
richardyz98: yes I do think that is a good idea

cassiopeia/solver/GreedySolver.py
mattjones315: Cool, let's change this function to `transform_priors` and here we can just pass in -1*np.log10.

cassiopeia/solver/dissimilarity_functions.py
mattjones315: Okay, let's keep it for now but can you change the name to `hamming_distance_without_missing` or something so a user is not confused?

cassiopeia/solver/solver_utilities.py
mattjones315: Sweet, let's make this change and then we can merge this code in.

cassiopeia/solver/MaxCutGreedySolver.py
```diff
@@ -92,51 +95,88 @@ def perform_split(
...
             return samples, []
 
         left_set = []
         right_set = []
         missing = []
 
         for i in samples:
-            if self.prune_cm.iloc[i, char] == state:
+            if (
+                self.unique_character_matrix.loc[i, :][chosen_character]
```
mattjones315, line 146: Why not index like this: `self.unique_character_matrix.loc[i, chosen_character]`?

cassiopeia/solver/MaxCutGreedySolver.py
```diff
@@ -92,51 +95,88 @@ def perform_split(
...
+                == chosen_state
+            ):
                 left_set.append(i)
-            elif self.prune_cm.iloc[i, char] == self.missing_char:
+            elif (
+                self.unique_character_matrix.loc[i, :][chosen_character]
```
mattjones315, line 151: Same as above.

cassiopeia/solver/SpectralGreedySolver.py
```diff
@@ -50,33 +56,54 @@ class SpectralGreedySolver(GreedySolver.GreedySolver):
...
+        )
 
         self.missing_data_classifier = missing_data_classifier
         self.threshold = threshold
-        self.weights = weights
+        if similarity_function:
+            self.similarity_function = similarity_function
+        else:
+            self.similarity_function = (
+                dissimilarity_functions.hamming_similarity_without_missing
```
mattjones315, line 99: If we default to `hamming_distance_without_missing`, can you make this clear in the documentation of the class?

cassiopeia/solver/SpectralGreedySolver.py
```diff
@@ -94,53 +121,89 @@ def perform_split(
...
             return samples, []
 
         left_set = []
         right_set = []
         missing = []
 
         for i in samples:
-            if self.prune_cm.iloc[i, char] == state:
+            if (
+                self.unique_character_matrix.loc[i, :][chosen_character]
```
mattjones315, line 172: Indexing issue.

cassiopeia/solver/SpectralGreedySolver.py
```diff
@@ -94,53 +121,89 @@ def perform_split(
...
+                == chosen_state
+            ):
                 left_set.append(i)
-            elif self.prune_cm.iloc[i, char] == self.missing_char:
+            elif (
+                self.unique_character_matrix.loc[i, :][chosen_character]
```
mattjones315, line 177: Indexing issue.

cassiopeia/solver/VanillaGreedySolver.py
```diff
@@ -78,40 +85,72 @@ def perform_split(
...
         for i in samples:
-            if self.prune_cm.iloc[i, char] == state:
+            if (
+                self.unique_character_matrix.loc[i, :][chosen_character]
+                == chosen_state
+            ):
                 left_set.append(i)
-            elif self.prune_cm.iloc[i, char] == self.missing_char:
+            elif (
+                self.unique_character_matrix.loc[i, :][chosen_character]
```
mattjones315, line 141: Indexing.

cassiopeia/solver/VanillaGreedySolver.py
```diff
@@ -78,40 +85,72 @@ def perform_split(
...
             return samples, []
 
         left_set = []
         right_set = []
         missing = []
 
         for i in samples:
-            if self.prune_cm.iloc[i, char] == state:
+            if (
+                self.unique_character_matrix.loc[i, :][chosen_character]
```
mattjones315, line 136: Indexing.

mattjones315 (commented):
Just some small clerical issues here to resolve.

cassiopeia/solver/MaxCutGreedySolver.py
richardyz98: this actually throws an error if the column indices aren't explicitly integers, so this way allows you to use the indexing of the row by index and the column by integer

cassiopeia/solver/SpectralGreedySolver.py
richardyz98: I think this is a working default for now, but I'd like to benchmark different similarities

cassiopeia/solver/MaxCutGreedySolver.py
richardyz98: kind of like a hybrid of loc for row and iloc for column

mattjones315: Sounds good, thanks for answering!

cassiopeia/solver/SpectralGreedySolver.py
mattjones315: Can you put it in the documentation for now, and leave a TODO for experimenting to find the best default dissimilarity function?

richardyz98: Yes will do

mattjones315 (approved):
Looks good, thanks!

PR #46, opened 2020-12-11
ILP Solver

mattjones315:
Implemented an ILP solver. Cythonized to the best of my ability some of the slower procedures (specifically the inference of a given potential graph layer). Tests can be found in the test directory.

Also added a new Makefile that lets you install with a simple call to `make install` via command line. You can also run all tests using `make test` from command line.


cassiopeia/solver/ILPSolver.py
```diff
@@ -0,0 +1,557 @@
...
+        Given a potential graph, a root to treat as the source, and a list of
+        targets, create a Gurobi mixed integer linear programming instance for
+        computing the Steiner Tree.
+
+        Args:
+            potential_graph: Potential graph representing the evolutionary
+                space on which to solve for the Steiner Tree.
+            root: A node in the graph to treat as the source.
+            targets: A list of nodes in the tree that serve as targets for the
+                Stiner Tree procedure.
```
richardyz98, line 317: "Steiner"

cassiopeia/solver/ILPSolver.py
```diff
@@ -0,0 +1,557 @@
...
+
+        Args:
+            potential_graph: Potential graph representing the evolutionary
+                space on which to solve for the Steiner Tree.
+            root: A node in the graph to treat as the source.
+            targets: A list of nodes in the tree that serve as targets for the
+                Stiner Tree procedure. 
+        
+        Returns:
+            A Gurobipy Model intance and the edge variables involved.
```
richardyz98, line 320: "instance"

cassiopeia/solver/ilp_solver_utilities.pyx
```diff
@@ -0,0 +1,102 @@
...
+    for i in range(0, n_samples-1):
+
+        sample1 = source_nodes[i]
+        top_ancestors = []
+        distance_to_ancestors = []
+
+        for j in range(i + 1, n_samples):
+
+            sample2 = source_nodes[j]
+            ancestor = np.array([sample1[k] if sample1[k] == sample2[k] else 0 for k in range(dim)])
```
richardyz98, line 67: should we include the logic where if a missing value appears, we infer the ancestor as the non-missing value

richardyz98 (changes requested):

cassiopeia/solver/ILPSolver.py
```diff
@@ -0,0 +1,557 @@
...
+            Ignored if set to 0.
+        convergence_iteration_limit: Number of iterations allowed for ILP
+            convergence. Ignored if set to 0.
+        maximum_potential_graph_layer_size: Maximum size allowed for an iteration
+            of the potential graph inference procedure. If this is exceeded, 
+            we return the previous iteration's graph or abort altogether.
+        weighted: Weight edges on the potential graph by the negative log
+            likelihood of the mutations.
+        seed: Random seed to use during ILP optimization.
+        mip_gap: Objective gap for mixed integer linear programming problem.
```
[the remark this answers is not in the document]
mattjones315: Great, thanks.

cassiopeia/solver/ILPSolver.py
```diff
@@ -0,0 +1,557 @@
...
+
+        # determine diameter of the dataset by evaluating maximum distance to
+        # the root from each sample
+        max_lca_distance = 0
+        lca_distances = [
+            dissimilarity_functions.hamming_distance(root, np.array(u))
+            for u in targets
+        ]
+        for i in range(len(lca_distances)):
+            for j in range(i + 1, len(lca_distances)):
```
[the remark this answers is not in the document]
mattjones315: Thanks for the tip, it's changed now.

cassiopeia/solver/ilp_solver_utilities.pyx
mattjones315: Yeah this is a great point - I've added this now. I'm just using your `get_lca_characters` function.

cassiopeia/solver/ILPSolver.py
```diff
@@ -0,0 +1,557 @@
...
+                n
+                for n in processed_solution
+                if processed_solution.in_degree(n) == 0
+            ]
+
+        processed_solution = self.append_sample_names(processed_solution)
+
+        return processed_solution
+
+    def append_sample_names(self, solution: nx.DiGraph) -> nx.DiGraph:
```
richardyz98, line 536: So if I understand it, this function appends the samples as explicit nodes into the graph? So this will append all duplicate samples and also append explicit sample nodes such that no sample is an internal node in the final tree?

mattjones315: yes, the purpose of this function is to append sample names onto their appropriate character states in the tree was just built. 

To reduce the complexity, we build the tree over unique character states. So we need to append sample names appropriately to the tree after we've built it for interpretability.

mattjones315: Ah I realized I hadn't documented this function - I'll add documentation.

richardyz98: Thanks Matt!

cassiopeia/solver/MaxCutSolver.py
```diff
@@ -3,8 +3,7 @@
 implements an inference procedure inspired by Snir and Rao (2006) that 
 approximates the max-cut problem on a connectivity graph generated from the 
 observed mutations on a group of samples. The connectivity graph represents a
-supertree generated over phylogenetic trees for each individual character, and
```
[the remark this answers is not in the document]
mattjones315: Hmm. Okay I'll copy over the most up to date version in the `refactor` branch.

cassiopeia/solver/ILPSolver.py
```diff
@@ -0,0 +1,557 @@
...
+        objective_expression = gurobipy.quicksum(
+            edge_variables_binary[u, v] * potential_graph[u][v]["weight"]
+            for u, v in potential_graph.edges()
+        )
+        model.setObjective(objective_expression, gurobipy.GRB.MINIMIZE)
+
+        return model, edge_variables
+
+    def solve_steiner_instance(
+        self, model, edge_variables, potential_graph, pid
```
richardyz98, line 399: There is also a lack of argument documentation here

PR #53, opened 2021-01-20
Fix failing test: Rename mapping => character_state_mapping.keys

sprillo:


sprillo:
This was caused by a missing commit, which got pushed directly into `refactor`.

PR #55, opened 2021-01-20
Remove duplicated code on error raising.

sprillo:
Many of CassiopeiaTree's methods operate on the underlying tree, and raise an error if the __network doesn't exist. I factored out this error-checking code to reduce duplication.

### DIFF
--- a/cassiopeia/data/CassiopeiaTree.py
+++ b/cassiopeia/data/CassiopeiaTree.py
@@ -147,6 +147,10 @@ class CassiopeiaTree:
                 self.__network.nodes[u]["time"] + self.__network[u][v]["length"]
             )
 
+    def __check_network_initialized(self):
+        if self.__network is None:
+            raise CassiopeiaTreeError("Tree has not been initialized.")
+
     def initialize_character_states_at_leaves(
         self, character_matrix: Union[pd.DataFrame, Dict]
     ):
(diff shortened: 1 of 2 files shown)


cassiopeia/data/CassiopeiaTree.py
```diff
@@ -147,6 +147,10 @@ def populate_tree(self, tree: Union[str, ete3.Tree, nx.DiGraph]):
                 self.__network.nodes[u]["time"] + self.__network[u][v]["length"]
             )
 
+    def __check_network_initialized(self):
+        if self.__network is None:
+            raise CassiopeiaTreeError("Tree has not been initialized.")
+
```
mattjones315, line 153: I really like this -- one of the things I was thinking about doing last night, per your suggestion, was add a decorator to make these checks. I couldn't figure out how to do it, but perhaps this would be even cleaner.

I'm happy with this solution though.

mattjones315 (approved):
I think this looks great, no major comments. Just a wish that we could make it a decorator, as you suggested, because that would be cooler. But we can do that in a future commit - already this cuts down on a lot of duplicated code.

cassiopeia/data/CassiopeiaTree.py
sprillo: I figured that (1) a decorator doesn't cut down on the lines of code any more than this (both approaches just use 1 line of code); and (2) decorators are hackier (whereas the current code can be understood by any python novice). So my conclusion was: the problem I am trying to solve here is so simple that using a decorator is overkill.

mattjones315: Sounds good -- so pragmatic of you! :-). I'm good to go on this PR whenever Richard has an opportunity to review it.

richardyz98 (approved):

PR #56, opened 2021-01-26
Bugfix: do not expose internals of the CassiopeiaTree

sprillo:
E.g. `return self.__cache["leaves"][:]` instead of `return self.__cache["leaves"]`. (The latter returns a pointer to the internal cached list; if the user modifies it the cache breaks.)


mattjones315 (changes requested):
Looks great, and thanks for catching this bug!

Can you add a quick test to make sure that the internal objects are not being modified if you, for example, get the nodes and change the name (or delete a node)?

mattjones315 (approved):
Wonderful tests! I'll merge this in now.

PR #60, opened 2021-02-04
Process pipeline richard

richardyz98:
Hey Matt and Sebastian,
This is the latest PR with fixed included.
These include:
-  all clerical fixes regarding variable names etc.
-  Fixes to the maxcut tests
- Testing and implementation of priors in the average missing data imputation
- Changing perform_split to take strings and respecting the string sample interface
- Changing transform_priors to take a string indicating the transformation, and implementing a set of functions


cassiopeia/solver/SpectralGreedySolver.py
```diff
@@ -119,12 +123,14 @@ def perform_split(
...
-            samples: A list of samples, represented as integer indices
+            samples: A list of samples, represented by their string names
 
         Returns:
             A tuple of lists, representing the left and right partition groups
         """
-        mutation_frequencies = self.compute_mutation_frequencies(samples)
+        int_samples = list(map(lambda x: self.name_to_index[x], samples))
+
+        mutation_frequencies = self.compute_mutation_frequencies(int_samples)
```
mattjones315, line 133: I noted this above, but I think that `compute_mutation_frequencies` should take in a list of sample names, not integer indices. You can map them to integer indices within the function itself if you'd like, but I think it's more usable if we pass in string names.

cassiopeia/solver/MaxCutSolver.py
```diff
@@ -95,19 +105,21 @@ def perform_split(
...
-            samples: A list of samples, represented as integer indices
+            samples: A list of samples, represented by their string names
 
         Returns:
             A tuple of lists, representing the left and right partition groups
         """
-        mutation_frequencies = self.compute_mutation_frequencies(samples)
+        int_samples = list(map(lambda x: self.name_to_index[x], samples))
+
+        mutation_frequencies = self.compute_mutation_frequencies(int_samples)
```
mattjones315, line 115: ditto about string names should be passed to `compute_mutation_frequencies`

cassiopeia/solver/MaxCutGreedySolver.py
```diff
@@ -91,12 +98,14 @@ def perform_split(
...
-            samples: A list of samples, represented as integer indices
+            samples: A list of samples, represented by their string names
 
         Returns:
             A tuple of lists, representing the left and right partition groups
         """
-        mutation_frequencies = self.compute_mutation_frequencies(samples)
+        int_samples = list(map(lambda x: self.name_to_index[x], samples))
+
+        mutation_frequencies = self.compute_mutation_frequencies(int_samples)
```
mattjones315, line 108: ditto - string names passed to compute_mutation_frequencies

cassiopeia/solver/SpectralSolver.py
```diff
@@ -119,19 +123,20 @@ def perform_split(
...
         Returns:
             A tuple of lists, representing the left and right partition groups
         """
+        int_samples = list(map(lambda x: self.name_to_index[x], samples))
 
         G = graph_utilities.construct_similarity_graph(
             self.unique_character_matrix,
             self.missing_char,
-            samples,
+            int_samples,
```
mattjones315, line 136: Would it be better here to pass in actual sample names to `construct_similarity_graph`? 

I feel like generally we're actually making these function _less_ usable, in favor of efficiency. I personally think that the conversion to integer indices and numpy arrays should occur behind the scenes, all the time, so that regular users aren't confused about how to interface with these function. It's a similar point to how we'd like to interface with the `calculate_mutation_frequencies` function.

cassiopeia/solver/VanillaGreedySolver.py
```diff
@@ -81,12 +88,14 @@ def perform_split(
...
-            samples: A list of samples, represented as integer indices
+            samples: A list of samples, represented by their string names
 
         Returns:
             A tuple of lists, representing the left and right partition groups
         """
-        mutation_frequencies = self.compute_mutation_frequencies(samples)
+        int_samples = list(map(lambda x: self.name_to_index[x], samples))
+
+        mutation_frequencies = self.compute_mutation_frequencies(int_samples)
```
mattjones315, line 98: ditto - sample names to `compute_mutation_frequencies`.

cassiopeia/solver/dissimilarity_functions.py
```diff
@@ -52,22 +52,22 @@ def weighted_hamming_distance(
...
+                if weights:
+                    d += weights[i][s1[i]] + weights[i][s2[i]]
                 else:
                     d += 2
         else:
             if s1[i] != 0:
-                if w:
-                    d -= 2 * w[i][s1[i]]
+                if weights:
+                    d -= 2 * weights[i][s1[i]]
```
mattjones315, line 71: I actually think we should remove this functionality. IMO we should only be penalizing differences here (i.e., tallying dissimilarities). 

In the similarity functions this will obviously be present.

cassiopeia/solver/dissimilarity_functions.py
```diff
@@ -195,17 +167,17 @@ def weighted_hamming_similarity(
...
             else:
-                if not w:
+                if not weights:
                     d += 1
         else:
             if s1[i] != 0 and s2[i] != 0:
-                if w:
-                    d -= w[i][s1[i]] - w[i][s2[i]]
+                if weights:
+                    d -= weights[i][s1[i]] + weights[i][s2[i]]
```
mattjones315, line 180: same as above for the hamming_dissimilarity -- we should only be tallying similarities here, I think.

cassiopeia/solver/dissimilarity_functions.py
```diff
@@ -137,39 +137,11 @@ def hamming_distance(s1: np.array(int), s2: np.array(int)) -> int:
     return dist
```
mattjones315, line 137: All the other functions here return a score normalized by the number of non-missing characters. Do we want that behavior here? 

Or we can add an argument requesting the function normalize by the number of non-missing entries, or not?

cassiopeia/solver/graph_utilities.py
```diff
@@ -43,14 +43,14 @@ def construct_connectivity_graph(
     with distant mutations.
 
     Args:
-        cm: The character matrix of observed character states for all samples
+        character_matrix: The character matrix of observed character states for all samples
```
mattjones315, line 46: Noted above -- if this is a user-facing function, shall we make it a bit more usable and pass in an actual character matrix, with string names, that we convert to a numpy array within the function?

cassiopeia/solver/graph_utilities.py
```diff
@@ -166,15 +170,15 @@ def max_cut_improve_cut(G: nx.Graph, cut: List[int]) -> List[int]:
 
 
 def construct_similarity_graph(
-    cm: Dict,
+    character_matrix: np.array,
```
mattjones315, line 173: same comment about `character_matrix` with string names that gets converted within the function.

cassiopeia/solver/solver_utilities.py
```diff
@@ -174,22 +187,42 @@ def collapse_unifurcations(tree: ete3.Tree) -> ete3.Tree:
...
     Returns:
         A dictionary of weights for each character/state pair
     """
+    if prior_transformation not in [
+        "negative_log",
+        "negative_square_root_log",
+        "inverse",
+    ]:
+        logging.info("Please select one of the supported prior transformations")
+        raise PriorTransformationError()
```
mattjones315, line 217: I don't think this is the expected way to raise an exception -- just pass the string "Please select..." to the `PriorTransformationError`. No need to log, in my opinion.

cassiopeia/solver/solver_utilities.py
```diff
@@ -174,22 +187,42 @@ def collapse_unifurcations(tree: ete3.Tree) -> ete3.Tree:
...
+
+    prior_function = lambda x: -np.log(x)
+
+    if prior_transformation == "negative_square_root_log":
+        prior_function = lambda x: -np.log(np.sqrt(x))
```
mattjones315, line 222: Does this really make sense?

I had thought that it would be sqrt (1/x), and from the docstring I thought it was sqrt(-log(x)). I think when we discussed this earlier this week, it was sqrt(1/x).

cassiopeia/solver/solver_utilities.py
```diff
@@ -174,22 +187,42 @@ def collapse_unifurcations(tree: ete3.Tree) -> ete3.Tree:
...
+    if prior_transformation == "inverse":
+        prior_function = lambda x: 1 / x if x > 0 else 0
```
mattjones315, line 224: `x` is never going to be 0, I don't think.

And if it where, you'd have the same problem with the -log(x) transformation. You might just want to throw an error if priors of 0 are passed in.

test/solver_tests/dissimilarity_functions_test.py
```diff
@@ -27,7 +27,7 @@ def setUp(self):
         }
 
         self.weights = solver_utilities.transform_priors(
-            self.priors, lambda x: -np.log(x)
+            self.priors, "negative_log"
         )
```
mattjones315, line 31: can you add tests looking at the other prior transformations you added?

Also if there are errors that are expected to be thrown with certain inputs, can you please test that too?

test/solver_tests/hybrid_solver_test.py
```diff
@@ -170,14 +196,24 @@ def test_cutoff(self):
...
     def test_top_down_split_manual(self):
-        
+
         # test manually
-        mutation_frequencies = self.hybrid_pp_solver.top_solver.compute_mutation_frequencies(["a", "b", "c", "d", "e"])
+        mutation_frequencies = (
+            self.hybrid_pp_solver.top_solver.compute_mutation_frequencies(
+                [0, 1, 2, 3, 4]
+            )
+        )
```
mattjones315, line 205: This is a good example of why we'd like to have `compute_mutation_frequencies` take in sample names - this is pretty unreadable in my opinion. It seems much more natural to compute the frequencies for character states across samples of interest, and here it's difficult to know what these samples are.

mattjones315 (changes requested):
Lots of improvement here - thanks for the hard work!

I think there're still some things that need to be cleaned up though, can you take a look at my comments and address them as needed?

Thanks!

cassiopeia/solver/GreedySolver.py
```diff
@@ -156,14 +169,14 @@ def compute_mutation_frequencies(
...
+        sample set. Duplicate cells have their frequencies included. Generates
+        a dictionary that maps each character to a dictionary of state/sample
+        frequency pairs, allowing quick lookup.
 
         Args:
-            cm: The character matrix from which to calculate frequencies
-            missing_char: The character representing missing values
-            samples: The set of relevant samples in calculating frequencies
+            samples: The set of relevant samples in calculating frequencies,
+                represented by their string names
```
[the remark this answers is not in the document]
sprillo: To add to @mattjones315 's point, let's also make sure we get all those type hints and type references in the docstrings right.

cassiopeia/solver/dissimilarity_functions.py
richardyz98: Given that this is only used in the calculation of LCA distances in the potential graph for ILP, maybe we shouldn't normalize it?

mattjones315: I agree, let's not normalize it here. It's much easier to normalize after the fact than to un-normalize something.

cassiopeia/solver/GreedySolver.py
```diff
@@ -139,40 +135,43 @@ def _solve(samples):
...
-            pair
+            A dictionary containing frequency information for each character/
+            state pair
 
         """
-        subset_cm = self.unique_character_matrix[samples, :]
+        sample_indices = solver_utilities.convert_sample_names_to_indices(
+            self.unique_character_matrix.index, samples
+        )
+        subset_cm = self.unique_character_matrix.to_numpy()[sample_indices, :]
```
mattjones315, line 172: Just to confirm - the reason you are doing this is because it is a lot faster to to get entries from a numpy matrix with integer indices than to access a pandas dataframe with `.loc`, correct? 

Are we sure that the new function `convert_sample_names_to_indices` doesn't add the same overhead? Was there a reason why you went from having a `node_mapping` helper variable in the class?

cassiopeia/solver/missing_data_methods.py
```diff
@@ -2,74 +2,87 @@
...
+    for sample in missing_indices:
+        left_score = score_side(subset_character_array_left, sample)
+        right_score = score_side(subset_character_array_right, sample)
 
         if left_score / len(left_set) > right_score / len(right_set):
-            left_set.append(i)
+            left_set.append(indices[sample])
         else:
-            right_set.append(i)
+            right_set.append(indices[sample])
```
mattjones315, line 86: Totally fine logic, but a nitpicky comment -- you seem to use the terminology that "sample" is a name of a row in the original character matrix and index is the integer representation of this. But here, these nomenclatures are swapped, I believe. It makes the code a bit less readable (I would have imagined it would make more sense to say `left_set.append(samples[index])`, for example. Does that make sense?)

mattjones315 (commented):
Looking awesome! Two comments here that I'd like to address before merging it in. Thanks!

cassiopeia/solver/GreedySolver.py
mattjones315: I should have added -- I see later on how you rely on this `convert_sample_names_to_indices` function in contexts outside of a class, so I actually think it makes sense in general. Just making sure that we aren't creating extra lines of code here if it's the same overhead as `subset_cm = self.unique_character_matrix.lo[sample_indices].to_numpy()`

richardyz98: I did some runtime testing by using a jupyter notebook, and to my understanding the mapping to indices actually takes about as long as _one_ iloc or loc call. So I think it would indeed save us a lot of time, especially in the graph methods where iloc or loc would be called many many times. I saw this in benchmarking where I got order of magnitudes speedup

mattjones315: Fantastic - thanks for the clarification here. So indeed, `dataframe.to_numpy()[indices,:]` is a lot faster than `dataframe.loc[samples].to_numpy`. Interesting!

cassiopeia/solver/missing_data_methods.py
richardyz98: I agree here I'll fix this up!

cassiopeia/solver/GreedySolver.py
richardyz98: As far as the removing of the node_mapping attribute, I thought that keeping these attributes around might be confusing to a user who is looking at the class attributes, given that the class level is supposed to respect the string interface. I thought it might be nice just to have individual method handle its own conversion on the backend. If you feel that this isn't an issue and we could save some overhead, then I can change it back, but overall I think the runtime contribution of the conversions should be negligible.

mattjones315: I actually think it's a nice utility to have around and definitely cuts down on duplicate code. Good call!

richardyz98: Actually sorry to clarify, in this instance it's about the same of magnitude, so I will reduce the lines of code here.

mattjones315 (approved):

PR #61, opened 2021-02-09
Solver tree refactor

mattjones315:
Changing solver APIs to all ingest a CassiopeiaTree. The intended workflow is:

```
tree = cas.data.CassiopeiaTree(character_matrix, priors, etc.)
solver = cas.solver.DerivedCassiopeiaSolver(params)
solver.solve(tree)

newick = tree.get_newick()
...
```


cassiopeia/solver/DistanceSolver.py
```diff
@@ -24,122 +25,74 @@ class DistanceSolverError(Exception):
...
-
-    def solve(self):
+    def solve(
+        self,
+        cassiopeia_tree: CassiopeiaTree,
+        dissimilarity_map: Optional[pd.DataFrame] = None,
+        root_sample: Optional[str] = None,
+        root_tree: bool = False,
+        prior_transformation: str = "negative_log",
+    ) -> None:
```
sprillo, line 42: What are the pros and cons of having the arguments `dissimilarity_map , root_sample, root_tree , prior_transformation` be part of the `solve` method instead of specifying them some other way, e.g. through `__init__` or the `CassiopeiaTree`? One con of having them be part of `solve` is that it violates the uniform `solve` interface we want. As it stands this is not a type error because the arguments are all optional, but this looks weird to me. I realize that `root_sample` is a `cassiopeia_tree`-specific notion, so it would be ridiculous to pass `root_sample` to `__init__` and then `cassiopeia_tree` to `solve`. I agree with this. I don't know yet what the best solution is. But to this end, let me ask: how are the `dissimilarity_map`, `root_sample`, `root_tree`  and `prior_transformation` created?

BTW this is not a red flag to me, I am happy to merge with this as-is but want to understand from a design perspective if we are going in the right direction.

cassiopeia/solver/DistanceSolver.py
```diff
@@ -167,13 +120,117 @@ def solve(self):
...
+        extracting the character matrix, creating the dissimilarity map if
+        needed, and setting up the "root" sample if the tree will be rooted.
+
+        Args:
+            cassiopeia_tree: Input CassiopeiaTree to `solve`.
+            dissimilarity_map: Dissimilarity map. If this is None, the function
+                will create it using the dissimilarity function specified.
+            root_sample: Sample to treat as the root.
+            root_tree: Whether or not to root the tree after inference.
+            prior_transformation: Function to apply to priors to create weights.
```
sprillo, line 152: This argument doesn't exist, it should be `weights`?

cassiopeia/solver/GreedySolver.py
```diff
@@ -47,50 +48,38 @@ class GreedySolver(CassiopeiaSolver.CassiopeiaSolver):
...
 
         Returns:
             A tuple of lists, representing the left and right partition groups
         """
         pass
 
-    def solve(self) -> nx.DiGraph:
+    def solve(self,
+            cassiopeia_tree: CassiopeiaTree,
+            prior_transformation: str = "negative_log"):
```
sprillo, line 82: Same as for `DistanceSolver`, but in this case it seems clearer to me: why not have `prior_transformation` be part of `__init__`? It is a hyperparameter that is not `cassiopeia_tree`-dependent.

cassiopeia/solver/CassiopeiaSolver.py
```diff
@@ -15,38 +15,16 @@ class CassiopeiaSolver(abc.ABC):
...
 
-        self.character_matrix = character_matrix
-        self.missing_char = missing_char
-        self.meta_data = meta_data
-        self.tree = None
-        self.priors = priors
+        self.prior_transformation = prior_transformation
 
     @abc.abstractmethod
     def solve(self):
```
sprillo, line 30: Should be `solve(self, cassiopeia_tree: CassiopeiaTree) -> None`

sprillo (commented):
Looks great, only would like to discuss whether we are happy adding subclass-specific parameters to `solve`.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -131,7 +131,7 @@ def populate_tree(self, tree: Union[str, ete3.Tree, nx.DiGraph]) -> None:
         for n in self.nodes:
             if (
                 self.__original_character_matrix is not None
-                and n in self.__original_character_matrix.index.values
+                and n in self.__original_character_matrix.index.tolist()
```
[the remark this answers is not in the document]
mattjones315: `.values` returns a np.array and `tolist` returns a python list, I believe. I was getting a warning I believe by iterating across a `np.array` with `for n in ...` so I changed to `tolist.`

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -688,6 +688,10 @@ def get_newick(self) -> str:
         """Returns newick format of tree."""
         return utilities.to_newick(self.__network)
 
+    def get_network(self) -> nx.DiGraph:
```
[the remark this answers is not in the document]
mattjones315: This is intended to return the networkx topology, though I could imagine a user wanting all the information as node attributes. Then again, why not just use the CassiopeiaTree?

What are your thoughts here on what this should return?

cassiopeia/solver/DistanceSolver.py
mattjones315: I think we're totally on the same page regarding this. It feels a bit funky to me as well, truth be told. The main argument I have for including these as arguments to the `solve` function is that I don't think they are general enough to the algorithm to be included in the __init__ statement.

A simple example is if you have a panel of character matrices you'd like to all reconstruct with the same `DistanceSolver` routine. They'd all have specific dissimilarity maps, priors, character matrices, etc. So it wouldn't make sense to pass things like a `dissimilairty_map` or sample to treat as a root (i.e., `root_sample`) to the __init__ statement because it wouldn't apply generally.

Of course, I anticipate you'd agree here - this was the motivation for feeding in a `CassiopeiaTree` object to the `solve` method. It can be argued that we should extend the `CassiopeiaTree` class to contain a `dissimilarity_map` and putative root sample for input into a `DistanceSolver`. I'm totally happy to do this if you think that would be cleaner. On this note, I also suppose we should have `root_tree` be an argument to the class itself -- this actually _is_ a parameter of the `DistanceSolver`, right? I guess a use-case to consider is if you are comparing the effects of different rooting strategies. These would constitute different solvers, I think (take for example UPGMA which by definition returns a rooted tree).

Regarding the `prior_transformation`, I'm actually quite unsure where's the best place to put this. It could arguably go into the CassiopeiaTree itself, but it might be out of sight for a user when turning to a solver. I prefer having it as a parameter to `solve` because it's a data-specific operation that you might want to specify for a particular tree, in a particular routine, but perhaps not be married to the tree.

cassiopeia/solver/GreedySolver.py
mattjones315: I touched on this in a previous response, but I feel that it's too data-specific to be married to a `CassiopeiaSolver`. But truly, I can see the merit of it being part of the `__init__` and would be happy to put it there if you feel strongly.

mattjones315: I touched on this in a previous reply, but I felt that `prior_transformation` was too data-specific to be married to a solver. But truly I see the merit of having it be part of the `__init__` so I'm happy to put it there if you feel strongly.

cassiopeia/solver/CassiopeiaSolver.py
mattjones315: I was under the impression if a method does not return anything we don't need to have `-> None`. Is this not technically correct?

sprillo: Ah, the highlight was the missing `cassiopeia_tree: CassiopeiaTree` arguent, not the `None` return value 😂 . You are probably right about `None`, I'm just paranoid and add it too, no need to fix that.

cassiopeia/data/CassiopeiaTree.py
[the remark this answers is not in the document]
sprillo: Ah, nice catch.

cassiopeia/data/CassiopeiaTree.py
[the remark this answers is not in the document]
sprillo: I agree, I would just commit to returning the tree topology. My only proposal is to perhaps rename this to `get_tree_topology` to be more explicit and the user to not have any surprises.

cassiopeia/solver/DistanceSolver.py
sprillo: Yes I think we are on the same page. Regarding data-specific objects that don't quite fit into the solver's `__init__`, and also not into the `CassiopeiaTree`, I'd only like to point out that the solver's `__init__` can alternatively take in a _function_ that computes the data-specific object. As a concrete example say that the `DistanceSolver` needs the root of the tree to be specified. Instead of:

```
class DistanceSolver(CassiopeiaSolver.CassiopeiaSolver):
    def __init__(
        self,
    ):
        ...

    def solve(tree: CassiopeiaTree, root: str):
        ...
```

and **also instead of**:

```
class DistanceSolver(CassiopeiaSolver.CassiopeiaSolver):
    def __init__(
        self,
        root: str
    ):
        ...

    def solve(tree: CassiopeiaTree):
        ...
```

one can have:
```
class DistanceSolver(CassiopeiaSolver.CassiopeiaSolver):
    def __init__(
        self,
        root_finding_algorithm: Callable[[CassiopeiaTree], str]
    ):
        ...

    def solve(tree: CassiopeiaTree):
        root = self.root_finding_algorithm(tree)
        ...
```

Anyway, the key is that this is the spirit of building objects by composition: The `DistanceSolver` becomes a composition of algorithms, one for specifying the root of the tree, one for computing prior transformations, etc. etc.

cassiopeia/data/CassiopeiaTree.py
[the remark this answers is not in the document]
mattjones315: Hmm. I think that's a bit unclear though since `get_network` connotes what you're expecting to get out. I'm open to changing it downstream - let's talk about it at our next standup!

cassiopeia/solver/DistanceSolver.py
mattjones315: This is a really cool suggestion! I'm not exactly sure how we would implement this, but I'll leave this as a TODO for the DistanceSolver class.

cassiopeia/data/CassiopeiaTree.py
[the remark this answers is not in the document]
sprillo: I don't have strong opinions about naming. I only think we should be clear about what is being returned: can they assume they'll get branch lengths, node times, etc. out of this? I think we should only guarantee the topology, even though it might have other attributes populated.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -769,3 +806,78 @@ def relabel_nodes(self, relabel_map: Dict[str, str]) -> None:
...
+
+    def set_dissimilarity_map(self, dissimilarity_map: pd.DataFrame):
+        """Sets the dissimilarity map variable in this object.
+        
+        Args:
+            dissimilarity_map: Dissimilarity map relating all N x N distances
+                between leaves.
+        """
+
+        self.__dissimilarity_map = dissimilarity_map.copy()
```
sprillo, line 824: If the keys in the `dissimilarity_map` don't match the keys in the `character_matrix`, when will the error be spotted?

cassiopeia/data/utilities.py
```diff
@@ -108,3 +111,47 @@ def _to_newick_str(g, node):
 
     root = [node for node in tree if tree.in_degree(node) == 0][0]
     return _to_newick_str(tree, root) + ";"
+
+def compute_dissimilarity_map(
+    cm: np.array,
+    C: int,
+    dissimilarity_function: Callable,
+    weights: Optional[Dict[int, Dict[int, float]]] = None,
+    missing_indicator: int = -1,
```
sprillo, line 120: This is called `missing_state_indicator` in the `CassiopeiaTree`, I think we should use consistent naming access the whole codebase.

cassiopeia/solver/DistanceSolver.py
```diff
@@ -167,42 +125,62 @@ def solve(self):
...
-    @numba.jit(forceobj=True)
-    def compute_dissimilarity_map(self, cm: np.array, C: int) -> np.array:
-        """Compute the dissimilarity between all samples
+        cassiopeia_tree.populate_tree(tree)
 
-        An optimized function for computing pairwise dissimilarities between
-        samples in a character matrix according to the dissimilarity function.
+    def setup_dissimilarity_map(
+        self, cassiopeia_tree: CassiopeiaTree
+    ) -> Tuple[pd.DataFrame, pd.DataFrame, Dict[str, List[str]], str]:
```
sprillo, line 136: I find the logic behind setting up the dissimilarity map tricky: Suppose the `cassiopeia_tree` already has a dissimilarity map in it, and the solver was initialized with a `dissimilarity_function`. Which of the two dissimilarity maps will prevail, and why? Might this surprise users?

cassiopeia/solver/DistanceSolver.py
```diff
@@ -250,7 +228,11 @@ def update_dissimilarity_map(
         """
         pass
 
-    def append_sample_names(self, solution: nx.DiGraph) -> nx.DiGraph:
+    def append_sample_names(
```
sprillo, line 231: Is this method private? (If so let's use an underscore)

cassiopeia/solver/HybridSolver.py
```diff
@@ -144,43 +165,58 @@ def solve(self):
...
         )
 
-        self.tree.add_node(root)
+        root = len(self.__tree.nodes) + 1
+
+        self.__tree.add_node(root)
         if len(clades) == 1:
             for clade in clades[0]:
-                self.tree.add_edge(root, clade)
+                self.__tree.add_edge(root, clade)
```
sprillo, line 219: Just curious: is there a reason to use double underscore vs single underscore for class attributes? Or is there no difference?

cassiopeia/solver/HybridSolver.py
```diff
@@ -233,52 +281,59 @@ def apply_bottom_solver(
...
+        subproblem_tree = subtree.get_network()
         subproblem_root = [
             n for n in subproblem_tree if subproblem_tree.in_degree(n) == 0
         ][0]
         subproblem_tree.add_edge(root, subproblem_root)
 
         return subproblem_tree, root
 
-    def assess_cutoff(self, samples: List[str]) -> bool:
+    def assess_cutoff(
```
sprillo, line 311: In general, which methods are public and which are private? I think it is important that we make this clear.

cassiopeia/data/CassiopeiaTree.py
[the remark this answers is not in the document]
mattjones315: This is a good point. I think you're right actually - I'll change the function to `get_tree_topology`. It would be too confusing, I think, to enumerate where exactly all of these data fields live in the networkx object being returned. 

I'll leave a TODO for us saying that it would be nice to have a function that will return a "rich" topology in networkx format with all of these attributes like branch lengths & node times.

cassiopeia/solver/DistanceSolver.py
mattjones315: The issue here is if the user wants to root the tree and has not specified a root sample -- then, we'll have to _add_ the root sample to the character matrix and recompute the dissimilarities according to the user-defined dissimilarity map. 

If the user _has_ a root sample, or does not desire to get a rooted tree, then we're good to go and nothing needs to be done (the dissimilarity map will not be overwritten and will be used as is).

cassiopeia/solver/DistanceSolver.py
mattjones315: Good point.

cassiopeia/solver/HybridSolver.py
mattjones315: I think the difference here is between "protected" and "private" attributes/functions. (https://www.tutorialsteacher.com/python/public-private-protected-modifiers)

I think variables that begin with a single underscore can be accessed publicly but not changed, whereas double underscores indicates that they should not be accessed publicly.

cassiopeia/data/CassiopeiaTree.py
mattjones315: This is a great question, & especially difficult to say what the best approach is because the CassiopeiaTree need not have both inside of it.

But, I'm also not sure they need to have the same samples. The only place where this kind of error could be introduced is in the `set_dissimilarity_map` function, so maybe we can just raise a Warning if a character matrix exists and the `dissimilarity_map` does not have the same samples.

richardyz98 (changes requested):

cassiopeia/solver/ILPSolver.py
```diff
@@ -71,7 +72,7 @@ def __init__(
         weighted: bool = False,
         seed: Optional[int] = None,
         mip_gap: float = 0.01,
-        prior_transformation: str = "negative_log"
+        prior_transformation: str = "negative_log",
     ):
 
         if weighted and not priors:
```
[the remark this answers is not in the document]
mattjones315: good catch!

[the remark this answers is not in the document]
mattjones315: Hmm. Actually, interestingly, this does not appear in my codebase. Maybe somehow these files are not up to date?

cassiopeia/data/utilities.py
mattjones315: Good point, I'll go through and change these.

cassiopeia/solver/DistanceSolver.py
sprillo: I see, there are some details involved that exceed my knowledge of the solver. I am cool with this, just wanted to make sure the logic is what we want!

cassiopeia/data/CassiopeiaTree.py
sprillo: I see, maybe some simple warning to catch an obvious mistake might be helpful? I don't know the details of the solver so I'll leave this as a suggestion, no need to implement it.

cassiopeia/solver/HybridSolver.py
sprillo: Ah, nice! Then double underscore is better. I didn't know about this!

cassiopeia/data/CassiopeiaTree.py
mattjones315: That's exactly what I did -- I throw a warning if a character matrix exists and the dissimilarity map does not have the same keys.

cassiopeia/solver/HybridSolver.py
mattjones315: I think anything not preceded by `__` will be public.

PR #62, opened 2021-02-09
Process pipeline richard

richardyz98:
Hi Matt and Sebastian,
This PR is everything Percolation. I also fixed a bug in the distancesolver, it wasn't saving the missing character as a class attribute. This is actually what was causing the poor performance in NJ.


cassiopeia/solver/PercolationSolver.py
```diff
@@ -0,0 +1,350 @@
...
+            cassiopeia_tree: CassiopeiaTree storing a character matrix and
+                priors.
+        """
+
+        # A helper function that builds the subtree given a set of samples
+        def _solve(
+            samples: List[Union[str, int]],
+            tree: nx.DiGraph,
+            character_matrix: pd.DataFrame,
+            unique_character_matrix: pd.DataFrame,
```
mattjones315, line 102: Is there a reason why you can't just calculate the `unique_character_matrix` from the `character_matrix`?

cassiopeia/solver/PercolationSolver.py
```diff
@@ -0,0 +1,350 @@
...
+            priors: Dict[int, Dict[int, float]],
+            weights: Dict[int, Dict[int, float]],
```
mattjones315, line 104: You could just pass the priors and compute the weights over and over again (not sure if this would rack up compute time or not)

cassiopeia/solver/PercolationSolver.py
```diff
@@ -0,0 +1,350 @@
...
+        ):
+            if len(samples) == 1:
+                return samples[0]
+            # Partitions the set of samples by percolating a similarity graph
+            clades = list(self.percolate(unique_character_matrix, samples, priors, weights, missing_state_indicator))
+            # Generates a root for this subtree with a unique int identifier
+            root = (
+                len(tree.nodes)
+                - unique_character_matrix.shape[0]
+                + character_matrix.shape[0]
```
mattjones315, line 115: Is there a reason why you don't set `root = len(tree.nodes)+1`?

cassiopeia/solver/PercolationSolver.py
```diff
@@ -0,0 +1,350 @@
...
+        weights = None
+        priors = None
+        if cassiopeia_tree.priors:
+            weights = solver_utilities.transform_priors(
+                cassiopeia_tree.priors, self.prior_transformation
+            )
+            priors = cassiopeia_tree.priors
+
+        # extract character matrix
+        character_matrix = cassiopeia_tree.get_original_character_matrix()
```
mattjones315, line 151: At other times you grab `cassiopeia_tree.get_current_character_matrix()` -- is there a reason you're using `original_character_matrix()` here?

cassiopeia/solver/PercolationSolver.py
```diff
@@ -0,0 +1,350 @@
...
+        unique_character_matrix = character_matrix.drop_duplicates()
+
+        tree = nx.DiGraph()
+        samples = list(unique_character_matrix.index)
+        for i in samples:
+            tree.add_node(i)
```
mattjones315, line 157: You can add all of these samples with one line of code: `tree.add_nodes_from(list(unique_character_matrix.index))`

cassiopeia/solver/PercolationSolver.py
```diff
@@ -0,0 +1,350 @@
...
+            A tuple of lists, representing the left and right partition groups
+        """
+        sample_indices = solver_utilities.convert_sample_names_to_indices(
+            character_matrix.index, samples
+        )
+        unique_character_array = character_matrix.to_numpy()
+
+        G = nx.Graph()
+        for v in sample_indices:
+            G.add_node(v)
```
mattjones315, line 205: nit: `G.add_nodes_from(sample_indices)`

cassiopeia/solver/PercolationSolver.py
```diff
@@ -0,0 +1,350 @@
...
+                successors = list(lca_network.successors(current_node))
+                if len(successors) == 1:
+                    current_node = successors[0]
+                else:
+                    for i in successors:
+                        grouped_components.append(
+                            solver_utilities.get_leaf_children(
+                                lca_network, i
+                            )
+                        )
```
mattjones315, line 284: instead of having this function that gets the leaves of a subtree in a network, why not use the function we already have in the `CassiopeiaTree` -- `lca_tree.leaves_in_subtree(i)`?

A similar argument can be made for using the CassiopeiaTree interface above for simple things like getting the root, but here I think it's especially worthwhile to mention that we don't need a new function in `solver_utilities` for this.

cassiopeia/solver/PercolationSolver.py
```diff
@@ -0,0 +1,350 @@
...
+            .to_dict()
+        )
+
+        for i in duplicate_groups:
+            if len(tree.nodes) == 1:
+                new_internal_node = len(duplicate_groups[i]) + 1
+            else:
+                new_internal_node = (
+                    max([n for n in tree.nodes if type(n) == int]) + 1
+                )
```
mattjones315, line 345: Same comment as above- why not use `new_internal_node = len(tree.nodes)+1`?

cassiopeia/solver/PercolationSolver.py
```diff
@@ -0,0 +1,350 @@
...
+            samples in the case that the percolation procedure generates more
+            than two groups of samples in the partition
+        prior_transformation: A function defining a transformation on the priors
+            in forming weights to scale the contribution of each mutation in
+            the similarity graph
+        similarity_function: A function that calculates a similarity score
+            between two given samples and their observed mutations. The default
+            is "hamming_distance_without_missing"
+        threshold: A minimum similarity threshold to include an edge in the
+            similarity graph
```
mattjones315, line 40: Can you give some intuition here, or somewhere else, what the similarity threshold is supposed to look like? (Like what does 1 mean? Is supposed to be between 0 and 1? etc.) What's a reasonable threshold? Is a good default really 0?

cassiopeia/solver/PercolationSolver.py
```diff
@@ -0,0 +1,350 @@
...
+                missing_state_indicator=missing_state_indicator,
+                priors=priors,
+            )
+
+            self.joining_solver.solve(lca_tree)
+            lca_network = lca_tree.get_tree_topology()
+            grouped_components = []
+            root = [
+                n for n in lca_network if lca_network.in_degree(n) == 0
+            ][0]
```
mattjones315, line 269: Nit: `lca_tree.root` will do this.

cassiopeia/solver/PercolationSolver.py
```diff
@@ -0,0 +1,350 @@
...
+
+            # Take the split at the root as the clusters of components
+            # in the split, ignoring unifurcations
+            current_node = root
+            while len(grouped_components) == 0:
+                successors = list(lca_network.successors(current_node))
```
mattjones315, line 275: Nit: `lca_tree.children(current_node)` will do this.

cassiopeia/solver/solver_utilities.py
```diff
@@ -258,3 +260,23 @@ def convert_sample_names_to_indices(
...
+
+    Returns:
+        A list of leaf descendants of a node
+    """
+    if tree.out_degree(node) == 0:
+        return [node]
+    leaves = []
+    for i in tree.successors(node):
+        leaves.extend(get_leaf_children(tree, i))
+    return leaves
```
mattjones315, line 282: Personally not a fan of reimplementing this function if we already have it in the `CassiopeiaTree` class. Or at least, let's  put in the `data.utilities` module since it's not specific to solvers.

test/solver_tests/percolation_test.py
```diff
@@ -0,0 +1,200 @@
...
+        ):
+            continue
+
+        if s1[i] == s2[i]:
+            if weights:
+                similarity += weights[i][s1[i]]
+            else:
+                similarity += 1
+
+    return -1 * similarity
```
mattjones315, line 56: Is there a reason why this is not in `dissimilarity_functions`? This looks like it could be useful.

test/solver_tests/percolation_test.py
```diff
@@ -0,0 +1,200 @@
...
+            ]
+        )
+        p_tree = cas.data.CassiopeiaTree(
+            cm, missing_state_indicator=-1
+        )
+
+        joining_solver = NeighborJoiningSolver(dissimilarity_function = neg_hamming_similarity_without_missing, add_root = True)
+        psolver = PercolationSolver(joining_solver = joining_solver)
+        psolver.solve(p_tree)
+        expected_newick_string = "((1,3),(2,0,(4,5)));"
```
mattjones315, line 79: I've actually found that perhaps a better way to test for structure is to create a new networkx graph with the correct structure and make sure that the triplets correct or RF or LCA structure is perfectly in agreement. The newick strings can be arbitrarily formed (for example - if you switched `(1,3)` to be `(3,1)` the structure would be the same).

test/solver_tests/percolation_test.py
```diff
@@ -0,0 +1,200 @@
...
+            (8, "c3"),
+            (10, 11),
+            (10, 12),
+            (11, "c7"),
+            (11, "c6"),
+            (12, "c4"),
+            (12, "c5"),
+        ]
+        for i in expected_edges:
+            self.assertIn(i, p_tree.get_tree_topology().edges)
```
mattjones315, line 120: Why not use `p_tree.edges`?

test/solver_tests/percolation_test.py
```diff
@@ -0,0 +1,200 @@
...
+            (8, "c1"),
+            (8, "c2"),
+            (8, "c3"),
+            (10, "c7"),
+            (10, 11),
+            (11, "c4"),
+            (11, "c5"),
+        ]
+        for i in expected_edges:
+            self.assertIn(i, p_tree.get_tree_topology().edges)
```
mattjones315, line 158: same comment as above - why not `p_tree.edges`?

test/solver_tests/percolation_test.py
```diff
@@ -0,0 +1,200 @@
...
+        joining_solver = NeighborJoiningSolver(dissimilarity_function = dissimilarity_functions.weighted_hamming_distance, add_root = True)
+        psolver = PercolationSolver(joining_solver = joining_solver)
+        psolver.solve(p_tree)
+        # Due to the way that networkx finds connected components, the ordering
+        # of nodes is uncertain
+        expected_newick_strings = [
+            "((c2,c4,(c5,c6)),(c1,c3));",
+            "((c2,c4,(c6,c5)),(c1,c3));",
+            "((c2,(c5,c6),c4),(c1,c3));",
+            "((c2,(c6,c5),c4),(c1,c3));",
```
mattjones315, line 194: If you set a seed, can you make this deterministic? Also - same comment as above, I think we may want to stay away from testing newick string equality.

mattjones315 (changes requested):
Awesome PR here! I left some comments and suggestions I'd like to get resolved before merging in. The big themes -- I think you actually have the opportunity to make use of the CassiopeiaTree class a bit more (especially with respect to the function for getting leaves in a subtree) and I think the specifics for the PercolationSolver can be a bit better documented for my understanding (e.g. the meaning behind the similarity threshold). This last point is actually something we can probably do a bit better job describing throughout the code base.

cassiopeia/solver/GreedySolver.py
```diff
@@ -136,7 +136,7 @@ def _solve(
             )
 
         # extract character matrix
-        character_matrix = cassiopeia_tree.get_original_character_matrix()
+        character_matrix = cassiopeia_tree.get_current_character_matrix()
```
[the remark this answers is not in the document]
richardyz98: I think we probably do want to use the current character matrix, as the user could for example infer missing data on the matrix beforehand and we would want to solve the post-processed CM

cassiopeia/solver/PercolationSolver.py
richardyz98: This is also currently how it's done in the GreedySolver, I'm guessing to cut down on the number of times we need to run df.drop_duplicates. Should I change the behavior in both? Running drop_duplicates doesn't seem to be a significant bottleneck

cassiopeia/solver/GreedySolver.py
```diff
@@ -225,9 +225,12 @@ def __add_duplicates_to_tree(
...
         for i in duplicate_groups:
-            new_internal_node = (
-                max([i for i in tree.nodes if type(i) == int]) + 1
-            )
+            if len(tree.nodes) == 1:
+                new_internal_node = len(duplicate_groups[i]) + 1
+            else:
+                new_internal_node = (
+                    max([n for n in tree.nodes if type(n) == int]) + 1
+                )
```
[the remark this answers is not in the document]
richardyz98: There is, which is that if we remove internal nodes in removing duplicates, then the length of tree.nodes could be reduced to be less than the value of an existing node. Basically this could cause a naming collision. I think this definitely is an issue with the way we're naming internal nodes and we should address this, but for now I included this hotfix

cassiopeia/solver/PercolationSolver.py
richardyz98: There isn't particularly, I'll change this

cassiopeia/solver/PercolationSolver.py
richardyz98: Sorry this is a typo

cassiopeia/solver/PercolationSolver.py
richardyz98: See above comment for the GreedySolver

cassiopeia/solver/GreedySolver.py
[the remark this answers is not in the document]
mattjones315: Gotcha. Yeah, it's a bit hacky but I understand why it needs to be there.

cassiopeia/solver/GreedySolver.py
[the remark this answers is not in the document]
mattjones315: That's a fair point. The question is if the user would already have a tree in the `CassiopeiaTree` object when we run it through a solver. If you take a look at the `set_character_matrix` code, we overwrite both character matrices. The only time the current and original character matrix would diverge is when states are set on particular leaves, and this can only happen if we have a tree already in the object.

All in all, this a good example of when having two types of character matrices is actually quite confusing!

cassiopeia/solver/PercolationSolver.py
mattjones315: Fair enough. It's all behind the scenes so it's not a big deal passing both in.

test/solver_tests/percolation_test.py
richardyz98: This is just the negative of the hamming_similarity_without_missing, but numba doesn't like lambda functions otherwise I would've used a lambda here

cassiopeia/solver/PercolationSolver.py
richardyz98: Sure I'll do that!

richardyz98: Actually I think I'll keep it in, as we pass it around too much and recalculating it at every step seems like it could be costly for large state distributions

test/solver_tests/percolation_test.py
mattjones315: Okay, I see. Then to reduce the amount of code - why not use the following code block?

```
def neg_hamming_similarity_without_missing(...):
     return -1 * hamming_similarity_without_missing(...)
```

richardyz98: Apparently numba doesn't like this either haha...

mattjones315: I got it to work - looks like you have to compile the dissimilarity function with numba first!

Reading more into it, it looks like that if you'd like to call a function from a numba function, then you need to make sure that function within is compiled with numba.

mattjones315 (approved):
Looks great! I'm going to merge it in.

PR #64, opened 2021-02-17
Move ILPSolver logfile argument from solve() to __init__()

sprillo:
As title. This aligns better with our design.


mattjones315:
Per #116  and our recent release, users can specify a logfile in `solve`, so this is no longer relevant.

PR #67, opened 2021-02-22
UPGMASolver

richardyz98:
Hi Matt and Sebastian,
This is the UPGMASolver PR. I changed some of the logic for the DistanceSolver as well, please let me know if you agree with it.

Sorry it's a bit of a long one with the tests, and thanks in advance for your comments and feedback!

### DIFF
--- a/cassiopeia/solver/NeighborJoiningSolver.py
+++ b/cassiopeia/solver/NeighborJoiningSolver.py
@@ -39,6 +39,13 @@ class NeighborJoiningSolver(DistanceSolver.DistanceSolver):
                 "square_root_inverse": Transforms each probability by the
                     the square root of 1/p
 
+    Attributes:
+        dissimilarity_function: Function used to compute dissimilarity between
+            samples.
+        add_root: Whether or not to add an implicit root the tree.
+        prior_transformation: Function to use when transforming priors into
+            weights.
+
     """
 
     def __init__(
@@ -157,7 +164,7 @@ class NeighborJoiningSolver(DistanceSolver.DistanceSolver):
             np.where(dissimilarity_map.index == cherry[1])[0][0],
         )
 
-        dissimilarity_array = self.update_dissimilarity_map_numba(
+        dissimilarity_array = self.__update_dissimilarity_map_numba(
             dissimilarity_map.to_numpy(), i, j
         )
         sample_names = list(dissimilarity_map.index) + [new_node]
@@ -177,10 +184,10 @@ class NeighborJoiningSolver(DistanceSolver.DistanceSolver):
 
     @staticmethod
     @numba.jit(nopython=True)
-    def update_dissimilarity_map_numba(
+    def __update_dissimilarity_map_numba(
         dissimilarity_map: np.array, cherry_i: int, cherry_j: int
     ) -> np.array:
-        """An optimized function for updating dissimilarities.
+        """A private, optimized function for updating dissimilarities.
 
         A faster implementation of updating the dissimilarity map for Neighbor
         Joining, invoked by `self.update_dissimilarity_map`.
--- a/cassiopeia/solver/NeighborJoiningSolver.py
+++ b/cassiopeia/solver/NeighborJoiningSolver.py
@@ -29,7 +29,7 @@ class NeighborJoiningSolver(DistanceSolver.DistanceSolver):
             map. Optional if a dissimilarity map is already provided.
         add_root: Whether or not to add an implicit root the tree, i.e. a root
             with unmutated characters. If set to False, and no explicit root is
-            provided in the CassiopeiaTree, then will return an unrooted, 
+            provided in the CassiopeiaTree, then will return an unrooted,
             undirected tree
         prior_transformation: Function to use when transforming priors into
             weights. Supports the following transformations:
@@ -58,9 +58,10 @@ class NeighborJoiningSolver(DistanceSolver.DistanceSolver):
             prior_transformation=prior_transformation,
         )
 
-
-    def root_tree(self, tree: nx.Graph, root_sample: str, remaining_samples: List[str]):
-        """Roots a tree at the inferred ancestral root.
+    def root_tree(
+        self, tree: nx.Graph, root_sample: str, remaining_samples: List[str]
+    ) -> nx.DiGraph():
+        """Roots a tree produced by Neighbor-Joining at the specified root.
 
         Uses the specified root to root the tree passed in
 
@@ -80,7 +81,6 @@ class NeighborJoiningSolver(DistanceSolver.DistanceSolver):
 
         return rooted_tree
 
-
     def find_cherry(self, dissimilarity_matrix: np.array) -> Tuple[int, int]:
         """Finds a pair of samples to join into a cherry.
 
@@ -98,16 +98,11 @@ class NeighborJoiningSolver(DistanceSolver.DistanceSolver):
         q = self.compute_q(dissimilarity_matrix)
         np.fill_diagonal(q, np.inf)
 
-        _min = np.argmin(q)
-
-        i, j = _min % q.shape[0], _min // q.shape[0]
-
-        return (i, j)
-
+        return np.unravel_index(np.argmin(q, axis=None), q.shape)
 
     @staticmethod
     @numba.jit(nopython=True)
-    def compute_q(dissimilarity_map: np.array(int)):
+    def compute_q(dissimilarity_map: np.array(int)) -> np.array:
         """Computes the Q-criterion for every pair of samples.
 
         Computes the Q-criterion defined by Saitou and Nei (1987):
@@ -140,7 +135,6 @@ class NeighborJoiningSolver(DistanceSolver.DistanceSolver):
         dissimilarity_map: pd.DataFrame,
         cherry: Tuple[str, str],
         new_node: str,
-        cluster_to_cluster_size: Dict[str, int] = None
     ) -> pd.DataFrame:
         """Update dissimilarity map after finding a cherry.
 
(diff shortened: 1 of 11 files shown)


cassiopeia/solver/UPGMASolver.py
```diff
@@ -0,0 +1,213 @@
...
+    ):
+
+        super().__init__(
+            dissimilarity_function=dissimilarity_function,
+            add_root=True,
+            prior_transformation=prior_transformation,
+        )
+
+
+    def root_tree(self, tree: nx.Graph, root_sample: str, remaining_samples: List[str]):
```
mattjones315, line 60: Should we make `root_tree` a private method? That way we don't need to really explain to users why some arguments are used or not used.

Generally, this would be cleaned up with Sebastian's suggestion of taking a compositional approach for proposing a root. Let's make a note of that!

cassiopeia/solver/UPGMASolver.py
```diff
@@ -0,0 +1,213 @@
...
+
+        Returns:
+            A tuple of integers representing rows in the dissimilarity matrix
+                to join.
+        """
+
+        dissimilarity_matrix = dissimilarity_matrix.astype(float)
+        np.fill_diagonal(dissimilarity_matrix, np.inf)
+
+        return np.unravel_index(np.argmin(dissimilarity_matrix, axis=None), dissimilarity_matrix.shape)
```
mattjones315, line 103: Is this better/faster than the way I find the `argmin` in `NeighborJoiningSolver`? If so, let's change the way I find the indices in `NeighborJoiningSolver`

cassiopeia/solver/UPGMASolver.py
```diff
@@ -0,0 +1,213 @@
...
+        Args:
+            dissimilarity_map: A dissimilarity map to update
+            cherry: A tuple of indices in the dissimilarity map that are joining
+            new_node: New node name, to be added to the new dissimilarity map
+
+        Returns:
+            A new dissimilarity map, updated with the new node
+        """
+        i_size, j_size = cluster_to_cluster_size[cherry[0]], cluster_to_cluster_size[cherry[1]]
+        cluster_to_cluster_size[new_node] = i_size + j_size
```
mattjones315, line 132: Yes, here I think you can use a private variable stored in the `UPGMASolver`. You can instantiate the mapping in the `__init__` function, I think.

cassiopeia/solver/UPGMASolver.py
```diff
@@ -0,0 +1,213 @@
...
+
+    @staticmethod
+    @numba.jit(nopython=True)
+    def update_dissimilarity_map_numba(
+        dissimilarity_map: np.array, cherry_i: int, cherry_j: int, size_i: int, size_j: int
+    ) -> np.array:
+        """An optimized function for updating dissimilarities.
+
+        A faster implementation of updating the dissimilarity map for Neighbor
+        Joining, invoked by `self.update_dissimilarity_map`.
```
mattjones315, line 165: "A faster implementation... for **UPGMA...**"

test/solver_tests/neighborjoining_solver_test.py
```diff
@@ -124,13 +124,57 @@ def delta_fn(
...
+            character_matrix=self.cm, dissimilarity_map=self.basic_dissimilarity_map, 
+        )
+
+        with self.assertRaises(Exception) as context:
+            nothing_solver.solve(no_root_tree)
 
+        self.assertTrue(
+            "Please specify an explicit root sample in the Cassiopeia Tree or specify the solver to add an implicit root"
+            in str(context.exception)
+        )
```
mattjones315, line 145: This is a dangerous test -- what if we change the error message? You can just leave it as `self.assertRaises(DistanceSolverException, nothing_solver.solve, no_root_tree)`

test/solver_tests/neighborjoining_solver_test.py
```diff
@@ -124,13 +124,57 @@ def delta_fn(
...
+        no_root_solver = cas.solver.NeighborJoiningSolver(
+            dissimilarity_function=None, add_root=True
+        )
+
+        with self.assertRaises(Exception) as context:
+            no_root_solver.solve(no_root_tree)
+
+        self.assertTrue(
+            "Please specify a dissimilarity function to add an implicit root, or specify an explicit root"
+            in str(context.exception)
```
mattjones315, line 156: ditto here, regarding the checking for errors thrown.

test/solver_tests/neighborjoining_solver_test.py
```diff
@@ -124,13 +124,57 @@ def delta_fn(
...
+            character_matrix=self.cm, root_sample_name = "b"
+        )
+
+        with self.assertRaises(Exception) as context:
+            nothing_solver.solve(root_only_tree)
+
+        self.assertTrue(
+            "Please specify a dissimilarity function or populate the CassiopeiaTree object with a dissimilarity map"
+            in str(context.exception)
+        )
```
mattjones315, line 169: ditto, checking for thrown errors

test/solver_tests/percolation_test.py
```diff
@@ -74,7 +74,7 @@ def test_NJ_negative_similarity(self):
         psolver = PercolationSolver(joining_solver=joining_solver)
         psolver.solve(p_tree)
 
-        expected_newick_string = "((1,3),(2,0,(4,5)));"
+        expected_newick_string = "((1,3),(0,2,(4,5)));"
```
mattjones315, line 77: This is a good example for why it's nice to check the __topology__ of a tree rather than the newick string - the ordering within is arbitrary, so we can get unstable tests.

test/solver_tests/neighborjoining_solver_test.py
```diff
@@ -124,13 +124,57 @@ def delta_fn(
...
+
+        nj_solver_fn = cas.solver.NeighborJoiningSolver(add_root=True, dissimilarity_function = delta_fn)
+        nj_solver_fn.solve(self.basic_tree)
+
+        self.assertEquals(self.basic_tree.get_dissimilarity_map().loc["a", "b"], 15)
```
mattjones315, line 174: This is giving me a warning -- can you change this to `self.assertEqual`?

test/solver_tests/upgma_test.py
```diff
@@ -0,0 +1,346 @@
...
+                sample2 = _leaves[j]
+                self.assertEqual(
+                    nx.shortest_path_length(T, sample1, sample2),
+                    nx.shortest_path_length(expected_tree, sample1, sample2),
+                )
+
+
+    def test_upgma_solver_weights(self):
+
+        self.upgma_solver_modified.solve(self.pp_tree_priors)
```
mattjones315, line 245: How stable are these results? Do you need to set a random state or seed?

mattjones315 (changes requested):
Looks fantastic Richard! You obviously learned the algorithm well, and I'm impressed how quickly!

I left a few comments regarding design of the `setup_implicit_root` functionality and the usage of the `cluster_to_cluster_sizes` variable. Can you address my comments before we merge this in?

I also had a few comments regarding some testing best practices -- 

(1) First, I think we should move to strictly testing the tree structure rather than the newick string since order can be a bit arbitrary.
(2) Second, I think testing for exact matches between error strings thrown by `DistanceSolverError`s is a bit dangerous because these error messages might change. It's better just to make sure we're throwing an error.

cassiopeia/solver/UPGMASolver.py
```diff
@@ -0,0 +1,213 @@
+"""
+This file stores a subclass of DistanceSolver, UPGMA. The inference procedure is
+the Neighbor-Joining algorithm proposed by Sokal and Michener (1958) that 
+iteratively joins together samples with the minimum dissimilarity.
+"""
```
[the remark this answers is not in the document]
richardyz98: Whoops sometimes when I move fast and miss these things. Thanks Matt!

cassiopeia/solver/UPGMASolver.py
richardyz98: Whoops thanks again!

cassiopeia/solver/NeighborJoiningSolver.py
```diff
@@ -55,6 +58,29 @@ def __init__(
...
+        """
+        tree.add_edge(remaining_samples[0], remaining_samples[1])
+
+        rooted_tree = nx.DiGraph()
+        for e in nx.dfs_edges(tree, source=root_sample):
+            rooted_tree.add_edge(e[0], e[1])
+
+        return rooted_tree
+
+
```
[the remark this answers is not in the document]
richardyz98: Yes I think this is something that should try to implement, but let's table it as a TODO for now!

cassiopeia/solver/UPGMASolver.py
richardyz98: So I think that inheritance with private methods is throwing me some errors, so I left it as a public method for now

cassiopeia/solver/UPGMASolver.py
```diff
@@ -0,0 +1,233 @@
...
+            columns=[cherry[0], cherry[1]],
+            index=[cherry[0], cherry[1]],
+            inplace=True,
+        )
+
+        return dissimilarity_map
+
+    @staticmethod
+    @numba.jit(nopython=True)
+    def update_dissimilarity_map_numba(
```
mattjones315, line 177: Should we make this function private? (If so, we should make it private in the `NeighborJoiningSolver` as well)

cassiopeia/solver/VanillaGreedySolver.py
```diff
@@ -53,9 +53,10 @@ def __init__(
         self,
         missing_data_classifier: Callable = missing_data_methods.assign_missing_average,
         prior_transformation: str = "negative_log",
+        collapse_tree: bool = True
```
mattjones315, line 56: Please add documentation for this `collapse_tree` argument, as with the other `GreedySolver` (sub)classes

mattjones315 (changes requested):
Great work here Richard! I have a few comments I'd like to get resolved -- 

(1) There are some documentation issues that I think need to be improved before merging in (mostly around some arguments not being explained)
(2) There is a function here and there where I think the logic can be cleaner, I've marked them where needed.

I've also made some of my own small changes that I'll push into this PR, around documentation. One thing I've noticed is that you have a `Returns` section even when there is nothing returned. I think it's best practice to only have `Returns` sections in documentation when something is actually returned. On this note, it would also be nice to have it clearly marked when an error is raised with a `Raises` section of the documentation (e.g. in `setup_dissimilarity_map`). I've added a `Raises` section for this function in particular, but keep this in mind going forward.

mattjones315 (approved):
Looks fantastic! I approve this for merging into refactor.

PR #69, opened 2021-02-25
Add simulation APIs

sprillo:
Let's make sure to discuss the package layout. I've places both `TreeSimulator` and `DataSimulator` under the same `simulator` package, but one can imagine other alternatives.

### DIFF
--- a/cassiopeia/simulator/DataSimulator.py
+++ b/cassiopeia/simulator/DataSimulator.py
@@ -1,0 +1,19 @@
+"""
+Abstract class DataSimulator, for overlaying data onto a CassiopeiaTree.
+
+All data simulators are derived classes of this abstract class, and at a minimum
+implement a method called `overlay_data`.
+"""
+import abc
+from cassiopeia.data import CassiopeiaTree
+
+
+class DataSimulator(abc.ABC):
+    """
+    DataSimulator is an abstract class that all data overlayers derive from.
+    """
+
+    @abc.abstractmethod
+    def overlay_data(self, tree: CassiopeiaTree) -> None:
+        """Overlay data on a CassiopeiaTree (in-place)"""
+        pass
--- a/cassiopeia/simulator/DataSimulator.py
+++ b/cassiopeia/simulator/DataSimulator.py
@@ -5,15 +5,30 @@ All data simulators are derived classes of this abstract class, and at a minimum
 implement a method called `overlay_data`.
 """
 import abc
+
 from cassiopeia.data import CassiopeiaTree
 
 
 class DataSimulator(abc.ABC):
     """
     DataSimulator is an abstract class that all data overlayers derive from.
+
+    A DataSimulator is very generic and meant to give users the flexibility to
+    overlay any kind of data onto the tree using this single API. The prime
+    example of data a user might want to overlay on a tree is lineage tracing
+    data, for which there is a specific subclass LineageTracingDataSimulator.
+    Other data of interest might include: transcriptomes, proteomes, etc.
     """
 
     @abc.abstractmethod
     def overlay_data(self, tree: CassiopeiaTree) -> None:
-        """Overlay data on a CassiopeiaTree (in-place)"""
+        """
+        Overlay data on a CassiopeiaTree (in-place).
+
+        The tree topology must be initialized.
+
+        Args:
+            tree: the CassiopeiaTree to overlay the data on. The tree topology
+                must be initialized.
+        """
         pass
--- a/cassiopeia/simulator/LineageTracingDataSimulator.py
+++ b/cassiopeia/simulator/LineageTracingDataSimulator.py
@@ -2,7 +2,7 @@
 This file stores an abstract subclass of DataSimulator, the
 LineageTracingDataSimulator. A LineageTracingDataSimulator overlays lineage
 tracing data onto a CassiopeiaTree, i.e. it sets the character states of a
-CassiopeiaTree.
+CassiopeiaTree (in particular, the character matrix).
 """
 import abc
 from cassiopeia.data import CassiopeiaTree
@@ -19,6 +19,7 @@ class LineageTracingDataSimulator(DataSimulator):
     def overlay_data(self, tree: CassiopeiaTree) -> None:
         """
         Overlay lineage tracing data onto the CassiopeiaTree (in-place).
-        This sets the character states of all nodes in the tree.
+        This sets the character states of all nodes in the tree, as well
+        as the character matrix.
         """
         pass
--- a/cassiopeia/simulator/LineageTracingDataSimulator.py
+++ b/cassiopeia/simulator/LineageTracingDataSimulator.py
@@ -1,0 +1,24 @@
+"""
+This file stores an abstract subclass of DataSimulator, the
+LineageTracingDataSimulator. A LineageTracingDataSimulator overlays lineage
+tracing data onto a CassiopeiaTree, i.e. it sets the character states of a
+CassiopeiaTree.
+"""
+import abc
+from cassiopeia.data import CassiopeiaTree
+from cassiopeia.simulator.DataSimulator import DataSimulator
+
+
+class LineageTracingDataSimulator(DataSimulator):
+    """
+    LineageTracingDataSimulator is an abstract class that all lineage tracing
+    data simulators derive from.
+    """
+
+    @abc.abstractmethod
+    def overlay_data(self, tree: CassiopeiaTree) -> None:
+        """
+        Overlay lineage tracing data onto the CassiopeiaTree (in-place).
+        This sets the character states of all nodes in the tree.
+        """
+        pass
--- a/cassiopeia/simulator/LineageTracingDataSimulator.py
+++ b/cassiopeia/simulator/LineageTracingDataSimulator.py
@@ -5,6 +5,7 @@ tracing data onto a CassiopeiaTree, i.e. it sets the character states of a
 CassiopeiaTree (in particular, the character matrix).
 """
 import abc
+
 from cassiopeia.data import CassiopeiaTree
 from cassiopeia.simulator.DataSimulator import DataSimulator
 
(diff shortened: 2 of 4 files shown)


cassiopeia/simulator/LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    LineageTracingDataSimulator is an abstract class that all lineage tracing
+    data simulators derive from.
+    """
+
+    @abc.abstractmethod
+    def overlay_data(self, tree: CassiopeiaTree) -> None:
+        """
+        Overlay lineage tracing data onto the CassiopeiaTree (in-place).
+        This sets the character states of all nodes in the tree.
```
richardyz98, line 22: Should we also specify that it populates the character matrix?

richardyz98 (changes requested):

cassiopeia/simulator/LineageTracingDataSimulator.py
sprillo: Thanks, added!

richardyz98 (approved):

cassiopeia/simulator/DataSimulator.py
```diff
@@ -0,0 +1,19 @@
+"""
+Abstract class DataSimulator, for overlaying data onto a CassiopeiaTree.
+
+All data simulators are derived classes of this abstract class, and at a minimum
+implement a method called `overlay_data`.
+"""
+import abc
+from cassiopeia.data import CassiopeiaTree
```
mattjones315, line 8: nit: style -> add new line between `import abc` and `from cassiopeia.data import CassiopeiaTree`

cassiopeia/simulator/DataSimulator.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class DataSimulator(abc.ABC):
+    """
+    DataSimulator is an abstract class that all data overlayers derive from.
+    """
+
+    @abc.abstractmethod
+    def overlay_data(self, tree: CassiopeiaTree) -> None:
+        """Overlay data on a CassiopeiaTree (in-place)"""
```
mattjones315, line 18: can you document this function a bit more? What do we expect to be in `tree`, what do we not expect? Will this modify the tree in place? etc.

cassiopeia/simulator/DataSimulator.py
```diff
@@ -0,0 +1,19 @@
...
+All data simulators are derived classes of this abstract class, and at a minimum
+implement a method called `overlay_data`.
+"""
+import abc
+from cassiopeia.data import CassiopeiaTree
+
+
+class DataSimulator(abc.ABC):
+    """
+    DataSimulator is an abstract class that all data overlayers derive from.
```
mattjones315, line 13: Let's expand a bit on the intended use case of the `DataSimulator` -- how should users use this, and how should it be extended? Are there certain things it should _not_ be used for?

cassiopeia/simulator/LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,25 @@
+"""
+This file stores an abstract subclass of DataSimulator, the
+LineageTracingDataSimulator. A LineageTracingDataSimulator overlays lineage
+tracing data onto a CassiopeiaTree, i.e. it sets the character states of a
+CassiopeiaTree (in particular, the character matrix).
+"""
+import abc
+from cassiopeia.data import CassiopeiaTree
+from cassiopeia.simulator.DataSimulator import DataSimulator
```
mattjones315, line 9: nit: style -> add new line between `import abc` and `from cassiopeia.data import CassiopeiaTree`

cassiopeia/simulator/LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class LineageTracingDataSimulator(DataSimulator):
+    """
+    LineageTracingDataSimulator is an abstract class that all lineage tracing
+    data simulators derive from.
```
mattjones315, line 15: Can you expand a bit more on what exactly "lineage tracing data" is? What should this populate in the `CassiopeiaTree`? When should users use this?

cassiopeia/simulator/LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+
+    @abc.abstractmethod
+    def overlay_data(self, tree: CassiopeiaTree) -> None:
+        """
```
mattjones315, line 20: Nit: Add a one line doc header

cassiopeia/simulator/LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,25 @@
...
+        Overlay lineage tracing data onto the CassiopeiaTree (in-place).
+        This sets the character states of all nodes in the tree, as well
+        as the character matrix.
```
mattjones315, line 23: Could you please add argument section to documentation?

Also, it would be great to expand a bit more on what we _expect_ trees to have and what they do not have. Should we also have a boolean argument controlling whether or not a user would receive a new tree or whether to perform the operation in place?

cassiopeia/simulator/TreeSimulator.py
```diff
@@ -0,0 +1,19 @@
...
+implement a method called `simulate_tree`.
+"""
+import abc
+from cassiopeia.data import CassiopeiaTree
+
+
+class TreeSimulator(abc.ABC):
+    """
+    TreeSimulator is an abstract class that all tree simulators derive from.
+    """
```
mattjones315, line 14: Can you define what a `TreeSimulator` is? If this is just a topology, we should make it clear. What is this intended to return? How are users supposed to interface with it? How would someone extend it?

cassiopeia/simulator/TreeSimulator.py
```diff
@@ -0,0 +1,19 @@
...
+
+    @abc.abstractmethod
+    def simulate_tree(self) -> CassiopeiaTree:
+        """Simulate a CassiopeiaTree"""
```
mattjones315, line 18: Could you please add a bit more documentation about the `simulate_tree` method? What considerations should a user have about this function, and what do they expect to get out of the function call? What fields will be populated in the `CassiopeiaTree`, and which ones would not be?

cassiopeia/simulator/__init__.py
```diff
@@ -0,0 +1,3 @@
+from DataSimulator import DataSimulator
+from LineageTracingDataSimulator import LineageTracingDataSimulator
+from TreeSimulator import TreeSimulator
```
mattjones315, line 3: what's the difference/benefit/caveat of using `from .DataSimulator import DataSimulator` vs `from DataSimulator import DataSimulator`?

mattjones315 (changes requested):

cassiopeia/simulator/DataSimulator.py
sprillo: Thanks!

cassiopeia/simulator/DataSimulator.py
sprillo: The docstring already specifies that the tree is modified in place (perhaps it wasn't clear?). I can't think about any more specific documentation that this at the moment, since baed on our discussion I believe that `DataSimulator` is a very generic class. Maybe we should remove it completely until we find a need for it? (right now only `LineageTracingDataSimulator` subclasses it, and is on its own abstract)?

cassiopeia/simulator/DataSimulator.py
sprillo: We should probably discuss whether we should have this class at all right now.

cassiopeia/simulator/LineageTracingDataSimulator.py
sprillo: I mentioned that the character states of the tree are set as well as the character matrix in the doscstring of the `overlay_data` method. Should these details be duplicated here in the class' docstring? A general question I have is what we want to say in the class' dosctring vs the method's docstring vs the module's dosctring.

cassiopeia/simulator/LineageTracingDataSimulator.py
sprillo: Thanks, totally forgot about the argument! I added that the trees should have a topology and meaningful branch lengths.

> Should we also have a boolean argument controlling whether or not a user would receive a new tree or whether to perform the operation in place?

I think this makes the interface less clear; all our interfaces currently operate on the tree in place. The users can make a copy of the tree if they want anyway?

cassiopeia/simulator/TreeSimulator.py
sprillo: Thanks, I've added a more lengthy description!

cassiopeia/simulator/TreeSimulator.py
sprillo: Added a minimal description, although I don't know how much to say here in `simulate_tree` vs in the class' docstring.

cassiopeia/simulator/__init__.py
sprillo: I'd want to believe there is no difference? If we've been using `from .X`  then let's use this for uniformity.

cassiopeia/simulator/DataSimulator.py
mattjones315: Your updated documentation is great -- this was what I was looking for. Just for the sake of posterity, to recap our conversation over slack, we'd like to have this `DataSimulator` abstract class because we'd like to extend the simulator to simulate spatial location, tissue identity, transcriptomic state, etc.

cassiopeia/simulator/TreeSimulator.py
```diff
@@ -5,15 +5,32 @@
...
 import abc
+
 from cassiopeia.data import CassiopeiaTree
 
 
 class TreeSimulator(abc.ABC):
     """
     TreeSimulator is an abstract class that all tree simulators derive from.
+
+    A TreeSimulator returns a CassiopeiaTree whose tree topology is initialized.
```
mattjones315, line 16: "whose tree topology is initialized" -- are we saying that a TreeSimulator should _not_ have branch lengths on the tree?

I see below we say that some `TreeSimulators` can return a tree with branch lengths -- why don't we make this slightly more clear: "A TreeSimulator returns a CassiopeiaTree with _at least_ its tree topology initialized"

mattjones315 (approved):
Looks great Sebastian, thanks for making these small adjustments to the documentation. I left some more comments but I think it's good to go whenever you'd like to merge it in.

cassiopeia/simulator/TreeSimulator.py
sprillo: Great suggestion!

PR #70, opened 2021-02-26
iTOL Plotting Library 

mattjones315:
I've added a new plotting library under `cassiopeia.plotting` (which can be accessed as `cassiopeia.pl`) which currently only contains functions for interfacing with iTOL.

Currently, the tests will skip any that require interfacing directly with iTOL if you do not have a config file (stored at `~/.itolconfig`). So you should not encounter errors if you do not have an iTOL account (to note, I was able to test the iTOL interface and it works fine, since I do have an iTOL account).

The only function exposed is an upload and export function, accessible via `cas.pl.upload_and_export_itol(...)`.

### DIFF
--- a/cassiopeia/pl.py
+++ b/cassiopeia/pl.py
@@ -1,0 +1,1 @@
+from cassiopeia.plotting import *
--- a/cassiopeia/plotting/itol_utilities.py
+++ b/cassiopeia/plotting/itol_utilities.py
@@ -6,8 +6,10 @@ on the iTOL software and how to create an account.
 
 import os
 
-from typing import Tuple, Optional
+import tempfile
+from typing import Dict, List, Tuple, Optional
 
+from bokeh import palettes
 from ete3 import Tree
 from itolapi import Itol
 from itolapi import ItolExport
@@ -26,19 +28,21 @@ class iTOLError(Exception):
 
     pass
 
+
 def upload_and_export_itol(
     cassiopeia_tree: CassiopeiaTree,
-    api_key: str,
-    projectName: str,
     tree_name: str,
     export_filepath: str,
+    itol_config: str = "~/.itolconfig",
+    api_key: Optional[str] = None,
+    projectName: Optional[str] = None,
     meta_data: List[str] = [],
     allele_table: Optional[pd.DataFrame] = None,
     indel_colors: Optional[pd.DataFrame] = None,
     indel_priors: Optional[pd.DataFrame] = None,
     rect: bool = False,
     include_legend: bool = False,
-    palette: List[str],
+    palette: List[str] = palettes.Category20[20],
 ):
     """Uploads a tree to iTOL and exports it.
 
@@ -47,7 +51,7 @@ def upload_and_export_itol(
     their tree. The function requires a user to have an account with iTOL and
     must pass in an `api_key` and a `project_name`, which corresponds to one
     in the user's iTOL account.
-    
+
     Args:
         cassiopeia_tree: A CassiopeiaTree instance, populated with a tree.
         api_key: API key linking to your iTOL account
@@ -69,7 +73,23 @@ def upload_and_export_itol(
         palette: A palette of colors in hex format.
     """
 
-    os.path.mkdir(".tmp/")
+    # create temporary direcotry with os.path.mktmp
+    temporary_directory = tempfile.mkdtemp()
+
+    if os.path.exists(itol_config):
+
+        with open(os.path.expanduser(itol_config), "r") as f:
+            config_string = f.read()
+
+        config = config.read_string(config_string)
+        api_key = config["DEFAULT"]["api_key"]
+        project_name = config["DEFAULT"]["project_name"]
+
+    if api_key is None or project_name is None:
+        raise iTOLError(
+            "Specify an api_key and project_name, or a valid iTOL "
+            "config file."
+        )
 
     with open(".tmp/tree_to_plot.tree", "w") as f:
         f.write(cassiopeia_tree.get_newick())
@@ -84,7 +104,6 @@ def upload_and_export_itol(
     itol_uploader = Itol()
     itol_uploader.add_file("tree_to_plot.tree")
 
-    
     files = []
     if allele_table is not None:
         files += create_indel_heatmap(
@@ -108,7 +127,7 @@ def upload_and_export_itol(
             )
 
         if pd.api.types.is_string_dtype(values):
-            colormap = palette[:len(values.unique())]
+            colormap = palette[: len(values.unique())]
             files += create_colorbar(
                 values,
                 cassiopeia_tree,
@@ -164,6 +183,7 @@ def upload_and_export_itol(
 
     # remove intermediate files
     os.remove("tree_to_plot.tree")
+    os.remove(temporary_directory)
 
 
 def create_gradient_from_df(
@@ -220,7 +240,7 @@ def create_colorbar(
     tree: CassiopeiaTree,
     colormap: Dict[str, Tuple[float, float, float]],
     dataset_name: str,
-    output_directory: str = "./.tmp/", 
+    output_directory: str = "./.tmp/",
     create_legend: bool = False,
 ):
 
@@ -369,9 +389,13 @@ def create_indel_heatmap(
             ]
 
         if len(str(j)) == 1:
-            alleleLabel_fileout = os.path.join(output_directory, f"indelColors_0{j}.txt")
+            alleleLabel_fileout = os.path.join(
+                output_directory, f"indelColors_0{j}.txt"
+            )
         elif len(str(j)) == 2:
-            alleleLabel_fileout = os.path.join(output_directory, f"indelColors_{j}.txt")
+            alleleLabel_fileout = os.path.join(
+                output_directory, f"indelColors_{j}.txt"
+            )
         with open(alleleLabel_fileout, "w") as ALout:
             for line in header:
                 ALout.write(line + "\n")
@@ -385,7 +409,10 @@ def create_indel_heatmap(
     return alfiles, rgb_heatmap
 
 
-def get_random_indel_colors(lineage_profile: pd.DataFrame) -> pd.DataFrame:
+def get_random_indel_colors(
+    lineage_profile: pd.DataFrame,
+    random_state: Optional[np.random.RandomState] = None,
+) -> pd.Series:
     """Assigns random color to each unique indel.
 
     Assigns a random HSV value to each indel observed in the specified
@@ -394,13 +421,14 @@ def get_random_indel_colors(lineage_profile: pd.DataFrame) -> pd.DataFrame:
     Args:
         lineage_profile: An NxM lineage profile reporting the indels observed
             at each cut site in a cell.
+        random_state: A random state for reproducibility
 
     Returns:
         A mapping from indel to HSV color.
     """
 
     unique_indels = np.unique(
-        lineage_profile.apply(lambda x: x.unique(), axis=0)
+        np.hstack(lineage_profile.apply(lambda x: x.unique(), axis=0))
     )
 
     # color families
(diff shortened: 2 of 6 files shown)


sprillo (changes requested):
I think you forgot to add `pl` directory? Tests fail with `ImportError: cannot import name 'pl' from 'cassiopeia'`.

mattjones315:
> I think you forgot to add `pl` directory? Tests fail with `ImportError: cannot import name 'pl' from 'cassiopeia'`.

mattjones315:
Hey @sprillo -- you found a total bonehead mistake of mine. I forgot to add the `pl.py` file to the PR! It's now up and hopefully will run smoothly now if you do `make install` and then `make test`

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,619 @@
...
+    shutil.rmtree(temporary_directory)
+
+
+def create_gradient_from_df(
+    df: pd.Series,
+    tree: CassiopeiaTree,
+    dataset_name: str,
+    output_directory: str = "./tmp/",
+    color_min: str = "#ffffff",
+    color_max: str = "#000000",
```
richardyz98, line 205: These fields are inaccesible from the outer function, should we add them to upload_and_export_itol

richardyz98 (changes requested):

cassiopeia/plotting/itol_utilities.py
richardyz98: Additionally, should we allow the user to specify min and max colors for the random color/indel generation?

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,619 @@
...
+                    )
+                )
+            else:
+                rgb_i = np.random.choice(range(len(colorlist)))
+                color_ranges = colorlist[rgb_i]
+                indel2color[indel] = rgb_to_hsv(
+                    generate_random_color(
+                        color_ranges[:2], color_ranges[2:4], color_ranges[4:6]
+                    )
+                )
```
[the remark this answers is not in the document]
mattjones315: Actually, I pass in `random_state` to the `generate_random_color` function in the `if` block.

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,619 @@
...
+
+    indel_colors = pd.DataFrame(columns=["color"])
+    for indel in indel2color.keys():
+        indel_colors.loc[indel, "color"] = indel2color[indel]
```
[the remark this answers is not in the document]
mattjones315: That's a great point -- I actually tried this as well.

However, there's something weird here. I'd like each entry in this dataframe to be a tuple `(r,g,b)` and if I create the pandas dataframe from dict or somehow else, it automatically converts the tuples into separate columns, which is not what I need.

In any event, I haven't noticed that this function is extremely slow. If it is we can think about alternative approaches of speeding things up.

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,619 @@
...
+    """
+
+    def assign_color(prob, random_state):
+        if random_state:
+            H = random_state.rand()
+        else:
+            H = np.random.rand()
+        S = prob
+        V = 0.5 + 0.5 * S
+        return (H, S, V)
```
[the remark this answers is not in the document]
mattjones315: Actually, these are HSV values -- they're an alternative encoding of colors. See here for more information: https://en.wikipedia.org/wiki/HSL_and_HSV

cassiopeia/plotting/itol_utilities.py
mattjones315: I don't think we should have th user specify min/max colors for the random color/indel generation -- it gets too hectic there. Plus, we already thought a lot about the random color generation a while ago. I think it works well.

I'd like to ideally pass in color bounds for the numerical data, but I don't see an elegant way of doing this. I think I should just add this as a todo.

cassiopeia/plotting/itol_utilities.py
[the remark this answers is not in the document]
richardyz98: I see that now thanks!

cassiopeia/plotting/itol_utilities.py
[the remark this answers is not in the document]
richardyz98: Thanks Matt!

cassiopeia/plotting/itol_utilities.py
richardyz98: That sounds like a plan

cassiopeia/plotting/itol_utilities.py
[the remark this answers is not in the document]
richardyz98: I see in that case let's proceed as is!

richardyz98 (approved):

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,623 @@
...
+    This function takes in a tree, plots in with iTOL, and exports it locally.
+    A user can also specify meta data and allele tables to visualize alongside
+    their tree. The function requires a user to have an account with iTOL and
+    must pass in an `api_key` and a `project_name`, which corresponds to one
+    in the user's iTOL account.
+
+    TODO(mgjones): Add the ability to pass in min/max colors for specific
+        numeric meta data to use when creating the gradient files.
+
+    Args:
```
sprillo, line 61: `itol_config` argument is missing? What is the expected structure of the config?

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,623 @@
...
+    # create temporary directory for storing files we'll upload to iTOL
+    temporary_directory = tempfile.mkdtemp()
+
+    if os.path.exists(os.path.expanduser(itol_config)):
+        
+        config = configparser.ConfigParser()
+        with open(os.path.expanduser(itol_config), "r") as f:
+            config_string = f.read()
+        config.read_string(config_string)
+        if api_key is None:
```
sprillo, line 92: There seems to be precedence for `api_key` and `project_name` over the `itol_config`, we should document that?

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,623 @@
...
+            api_key = config["DEFAULT"]["api_key"]
+        if project_name is None:
+            project_name = config["DEFAULT"]["project_name"]
+
+    if api_key is None or project_name is None:
```
sprillo, line 97: We can be a bit more conservative here and say: `if itol_config is None and (api_key is None or project_name is None)`; the current `if` condition relies on previous logic, which can potentially lead to bugs if the previous code changed.

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,623 @@
...
+    itol_uploader.params["APIkey"] = api_key
+    itol_uploader.params["projectName"] = project_name
+
+    good_upload = itol_uploader.upload()
+    if not good_upload:
+        raise iTOLError(itol_uploader.comm.upload_output)
+
+    print("iTOL output: " + str(itol_uploader.comm.upload_output))
+    print("Tree Web Page URL: " + itol_uploader.get_webpage())
+    print("Warnings: " + str(itol_uploader.comm.warnings))
```
sprillo, line 166: Should we toggle these print statements with a verbosity flag?

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,623 @@
...
+    else:
+        # circular tree settings
+        itol_exporter.set_export_param_value("display_mode", 2)
+        itol_exporter.set_export_param_value("arc", 359)
+        itol_exporter.set_export_param_value("rotation", 270)
+
+    itol_exporter.set_export_param_value("leaf_sorting", 1)
+    itol_exporter.set_export_param_value("label_display", 0)
+    itol_exporter.set_export_param_value("internal_marks", 0)
+    itol_exporter.set_export_param_value("ignore_branch_length", 1)
```
sprillo, line 187: How easy is it to plot the branch lengths or node ages too? Can we make this a flag?

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,623 @@
...
+        color_max: Maximum color for gradient
+
+    Returns:
+        The filepath to new gradient file.
+    """
+
+    _leaves = tree.leaves
+    df = df.loc[_leaves].copy()
+
+    outfps = []
```
sprillo, line 230: Unused variable.

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,623 @@
...
+            number_of_items = len(colormap)
+
+            SIDout.write(f"LEGEND_TITLE\t{dataset_name} legend\n")
+            SIDout.write("LEGEND_SHAPES")
+            for _ in range(number_of_items):
+                SIDout.write("\t1")
+
+            SIDout.write("\nLEGEND_COLORS")
+            for col in colormap.values():
+                SIDout.write(f"\t{rgb_to_hex(col)}")
```
sprillo, line 332: Where is `rgb_to_hex` defined?

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,623 @@
...
+                r[i, j], g[i, j], b[i, j] = 1, 1, 1
+            elif "none" in ind.lower():
+                r[i, j], g[i, j], b[i, j] = 192 / 255, 192 / 255, 192 / 255
+            else:
+                col = hsv_to_rgb(tuple(indel_colors.loc[ind, "color"]))
+                r[i, j], g[i, j], b[i, j] = col[0], col[1], col[2]
+
+    rgb_heatmap = np.stack((r, g, b), axis=2)
+
+    alfiles = []
```
sprillo, line 413: What does `alfiles` mean? Is it "all files"?

cassiopeia/plotting/itol_utilities.py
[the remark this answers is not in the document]
sprillo: do we prefer to explicitly call them `hue, saturation, value` though?

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,623 @@
...
+        "SHOW_INTERNAL\t0",
+        "",
+    ]
+
+    outfp = os.path.join(output_directory, f"{dataset_name}.txt")
+    with open(outfp, "w") as SIDout:
+        for line in header:
+            SIDout.write(line + "\n")
+
+        if create_legend:
```
sprillo, line 322: This logic is not covered by the tests.

sprillo (changes requested):
Looks great! Just left a couple comments.

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,623 @@
...
+
+import configparser
+import shutil
+import tempfile
+from typing import Dict, List, Tuple, Optional
+
+from bokeh import palettes
+from ete3 import Tree
+from itolapi import Itol
+from itolapi import ItolExport
```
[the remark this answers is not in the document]
mattjones315: Good call, I'd missed this.

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,623 @@
...
+
+import os
+
+import configparser
+import shutil
+import tempfile
+from typing import Dict, List, Tuple, Optional
+
+from bokeh import palettes
+from ete3 import Tree
```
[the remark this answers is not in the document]
mattjones315: Good catches.

cassiopeia/pl.py
```diff
@@ -0,0 +1 @@
+from cassiopeia.plotting import *
```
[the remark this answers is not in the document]
mattjones315: It's just slicker :-).

It's actually common nowadays to abbreviate the libraries (e.g. `preprocessing` -> `pp`, `plotting`->`pl`). It started with scanpy, but I think a lot of other packages are going this way.

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,623 @@
...
+
+
+class iTOLError(Exception):
+    """Raises errors related to iTOL plotting
+    """
+
+    pass
+
+
+def upload_and_export_itol(
```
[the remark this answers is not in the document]
mattjones315: No, because I think some more advanced users (including myself) might want to have a bit more control over the plotting and thus it's nice to expose these other functions if the user imports the entire file.

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,623 @@
...
+    indel_colors: Optional[pd.DataFrame] = None,
+    indel_priors: Optional[pd.DataFrame] = None,
+    rect: bool = False,
+    include_legend: bool = False,
+    palette: List[str] = palettes.Category20[20],
+    random_state: Optional[np.random.RandomState] = None,
+):
+    """Uploads a tree to iTOL and exports it.
+
+    This function takes in a tree, plots in with iTOL, and exports it locally.
```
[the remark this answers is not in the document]
mattjones315: Thanks for the catch.

cassiopeia/plotting/itol_utilities.py
```diff
@@ -0,0 +1,623 @@
...
+    A user can also specify meta data and allele tables to visualize alongside
+    their tree. The function requires a user to have an account with iTOL and
+    must pass in an `api_key` and a `project_name`, which corresponds to one
```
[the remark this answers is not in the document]
mattjones315: Nice catch in the documentation.

cassiopeia/plotting/itol_utilities.py
mattjones315: Done.

cassiopeia/plotting/itol_utilities.py
mattjones315: Good point. I've changed up the logic here to take this into account. Hope it makes sense.

cassiopeia/plotting/itol_utilities.py
mattjones315: done.

cassiopeia/plotting/itol_utilities.py
mattjones315: Yeah, this is a good idea -- I'll add an argument for taking into account branch lengths.

cassiopeia/plotting/itol_utilities.py
mattjones315: Thanks for catching this

cassiopeia/plotting/itol_utilities.py
mattjones315: Great catch.. I hadn't defined it! Goes to show why code coverage is so important...

cassiopeia/plotting/itol_utilities.py
mattjones315: No, these are "allele files". I'll clarify this.

cassiopeia/plotting/itol_utilities.py
mattjones315: Good point! I've covered this now.

PR #73, opened 2021-02-26
SharedMutationJoiner (UPGMA-like algorithm)

richardyz98:
Hi Matt and Sebastian,
Here is the SMJ (working title for the UPGMA-like algorithm).

I also made it so that the conversion from the numba calculated flattened array to the square nxn dissimilarity map happens inside the utilities.compute_dissimilarity_map function, as I thought it was more natural for it to handle this conversion.

Additionally, the function does not make use of the compute_dissimilarity_map function found in CassiopeiaTree as I felt we do not want to override the dissimilarity map in the Tree. This is due to the fact that the algorithm calculates a similarity map and not a dissimilarity map and this may cause some confusion. 

Please let me know what you think! Thanks in advance for your feedback.

### DIFF
--- a/cassiopeia/data/CassiopeiaTree.py
+++ b/cassiopeia/data/CassiopeiaTree.py
@@ -23,7 +23,6 @@ import ete3
 import networkx as nx
 import numpy as np
 import pandas as pd
-import scipy
 
 from cassiopeia.data import utilities
 from cassiopeia.solver import solver_utilities
@@ -897,7 +896,6 @@ class CassiopeiaTree:
             weights,
             self.missing_state_indicator,
         )
-        dissimilarity_map = scipy.spatial.distance.squareform(dissimilarity_map)
 
         dissimilarity_map = pd.DataFrame(
             dissimilarity_map,
--- a/cassiopeia/data/CassiopeiaTree.py
+++ b/cassiopeia/data/CassiopeiaTree.py
@@ -23,6 +23,7 @@ import ete3
 import networkx as nx
 import numpy as np
 import pandas as pd
+import scipy
 
 from cassiopeia.data import utilities
 from cassiopeia.solver import solver_utilities
@@ -896,6 +897,8 @@ class CassiopeiaTree:
             weights,
             self.missing_state_indicator,
         )
+
+        dissimilarity_map = scipy.spatial.distance.squareform(dissimilarity_map)
 
         dissimilarity_map = pd.DataFrame(
             dissimilarity_map,
--- a/cassiopeia/data/utilities.py
+++ b/cassiopeia/data/utilities.py
@@ -9,6 +9,7 @@ import numba
 import numpy as np
 import pandas as pd
 import re
+import scipy
 
 from cassiopeia.preprocess import utilities as preprocessing_utilities
 
@@ -138,7 +139,7 @@ def compute_dissimilarity_map(
         missing_state_indicator: State indicating missing data
 
     Returns:
-        A dissimilarity mapping as a flattened array.
+        An n x n pairwise dissimilarity matrix.
     """
 
     nb_dissimilarity = numba.jit(dissimilarity_function, nopython=True)
@@ -174,9 +175,8 @@ def compute_dissimilarity_map(
 
         return dm
 
-    return _compute_dissimilarity_map(
-        cm, C, missing_state_indicator, nb_weights
-    )
+    flat_dissimilarity_map = _compute_dissimilarity_map(cm, C, missing_state_indicator, nb_weights)
+    return scipy.spatial.distance.squareform(flat_dissimilarity_map)
 
 
 def sample_bootstrap_character_matrices(
--- a/cassiopeia/data/utilities.py
+++ b/cassiopeia/data/utilities.py
@@ -9,7 +9,6 @@ import numba
 import numpy as np
 import pandas as pd
 import re
-import scipy
 
 from cassiopeia.preprocess import utilities as preprocessing_utilities
 
@@ -139,7 +138,7 @@ def compute_dissimilarity_map(
         missing_state_indicator: State indicating missing data
 
     Returns:
-        An n x n pairwise dissimilarity matrix.
+        A dissimilarity mapping as a flattened array.
     """
 
     nb_dissimilarity = numba.jit(dissimilarity_function, nopython=True)
@@ -175,8 +174,9 @@ def compute_dissimilarity_map(
 
         return dm
 
-    flat_dissimilarity_map = _compute_dissimilarity_map(cm, C, missing_state_indicator, nb_weights)
-    return scipy.spatial.distance.squareform(flat_dissimilarity_map)
+    return _compute_dissimilarity_map(
+        cm, C, missing_state_indicator, nb_weights
+    )
 
 
 def sample_bootstrap_character_matrices(
--- a/cassiopeia/solver/SharedMutationJoiningSolver.py
+++ b/cassiopeia/solver/SharedMutationJoiningSolver.py
@@ -11,6 +11,7 @@ import networkx as nx
 import numba
 import numpy as np
 import pandas as pd
+import scipy
 
 from cassiopeia.data import CassiopeiaTree
 from cassiopeia.data import utilities as data_utilities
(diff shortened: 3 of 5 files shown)


cassiopeia/data/CassiopeiaTree.py
```diff
@@ -897,7 +896,6 @@ def compute_dissimilarity_map(
             weights,
             self.missing_state_indicator,
         )
-        dissimilarity_map = scipy.spatial.distance.squareform(dissimilarity_map)
```
mattjones315, line 900: I actually don't know if we should be doing this. Part of me thinks, actually, that we should be working with flattened dissimilarity matrices because they can get so large! (Think about a lineage with 1000 cells -- the square matrix will have 1000^2 = 1e6 entries when in reality half of those are duplicated).

cassiopeia/data/utilities.py
```diff
@@ -138,7 +139,7 @@ def compute_dissimilarity_map(
         missing_state_indicator: State indicating missing data
 
     Returns:
-        A dissimilarity mapping as a flattened array.
+        An n x n pairwise dissimilarity matrix.
```
mattjones315, line 142: To cut down on memory, I think we should be returning flattened arrays.

cassiopeia/data/utilities.py
```diff
@@ -174,9 +175,8 @@ def _compute_dissimilarity_map(cm, C, missing_state_indicator, nb_weights):
 
         return dm
 
-    return _compute_dissimilarity_map(
-        cm, C, missing_state_indicator, nb_weights
-    )
+    flat_dissimilarity_map = _compute_dissimilarity_map(cm, C, missing_state_indicator, nb_weights)
+    return scipy.spatial.distance.squareform(flat_dissimilarity_map)
```
mattjones315, line 179: ditto above

cassiopeia/solver/SharedMutationJoiningSolver.py
```diff
@@ -0,0 +1,298 @@
...
+    Implements an iterative, bottom-up agglomerative clustering procedure. The
+    algorithm iteratively clusters the samples in the sample pool by the number
+    of shared mutations that they have in their character information. The
+    algorithm has theoretical guarantees on correctness given a sufficiently
+    large number of characters and bounds on edge lengths in the tree generative
+    process.
+
+    Args:
+        dissimilarity_function: Function that can be used to compute the
+            dissimilarity between samples.
```
mattjones315, line 38: `dissimilarity_function` -> `similarity_function`

cassiopeia/solver/SharedMutationJoiningSolver.py
```diff
@@ -0,0 +1,298 @@
...
+        prior_transformation: Function to use when transforming priors into
+            weights. Supports the following transformations:
+                "negative_log": Transforms each probability by the negative
+                    log (default)
+                "inverse": Transforms each probability p by taking 1/p
+                "square_root_inverse": Transforms each probability by the
+                    the square root of 1/p
+
+    Attributes:
+        dissimilarity_function: Function used to compute dissimilarity between
```
mattjones315, line 48: ditto, re: naming of `dissimilarity_function`

cassiopeia/solver/SharedMutationJoiningSolver.py
```diff
@@ -0,0 +1,298 @@
...
+        ],
+        prior_transformation: str = "negative_log",
+    ):
+
+        super().__init__(prior_transformation)
+
+        self.similarity_function = similarity_function
+
+    def solve(self, cassiopeia_tree: CassiopeiaTree) -> None:
+        """The solver routine for the SharedMutationJoiningSolver.
```
mattjones315, line 67: Nit: `The solver...` -> `Solves a tree...`

I believe it's best code style to use verbs in the documentation titles.

cassiopeia/solver/SharedMutationJoiningSolver.py
```diff
@@ -0,0 +1,298 @@
...
+from cassiopeia.solver import CassiopeiaSolver, solver_utilities
+
+
+class SharedMutationJoiningSolverError(Exception):
+    """An Exception class for SharedMutationJoiningSolver."""
+
+    pass
+
+
+class SharedMutationJoiningSolver(CassiopeiaSolver.CassiopeiaSolver):
```
mattjones315, line 26: Is the only reason why we don't want to build on the `DistanceSolver` API because we don't want to overwrite the `dissimilarity_map` attribute in a tree?

The code in this class looks almost _identical_ to the code in the `DistanceSolver` classes, barring some name changes (e.g. dissimilarity -> similarity). If the only concern is that you don't want to overwrite the dissimilarity map, then there are three things we can do:

(1) Take your advice from last week and _not_ overwrite the dissimilarity map in a tree if you pass in a dissimilarity function.
(2) Copy the `CassiopeiaTree` where you can overwrite the dissimilarity map with a similarity map and then populate the original `CassiopeiaTree` with the resulting tree.
(3) Just run `utilities.compute_dissimilarity_map` on the character matrix in the original `CassiopeiaTree`. As opposed to `cassiopeia_tree.compute_dissimilarity_map`, this won't overwrite anything and will just get the dissimilarity map.

cassiopeia/solver/SharedMutationJoiningSolver.py
```diff
@@ -0,0 +1,298 @@
...
+
+        return similarity_map
+
+    @staticmethod
+    @numba.jit(nopython=True)
+    def __update_similarity_map_numba(
+        character_matrix: np.array,
+        similarity_map: np.array,
+        lca: np.array,
+        nb_similarity_function,
```
mattjones315, line 262: Missing typing for the `nb_similarity_function`

mattjones315 (changes requested):
Nice PR Richard!

I think though that you can use the `DistanceSolver` API here. This class almost looks identical to `DistanceSolver` and it'd be nice to cut down on duplicated code and it'd make the codebase more cohesive. I think I know why you don't want to just use the regular class API and I've offered some work arounds.

Let's hit some of my questions / comments / suggestions before merging this in.

cassiopeia/solver/SharedMutationJoiningSolver.py
richardyz98: I think that the main reason that I didn't want to use the DistanceSolver API is because in addition to just changing just the dissimilarity map like in UPGMA and in NJ, we also need to update and query a working character matrix as we introduce new LCA nodes from collapsing. I couldn't think of a good way to incorporate this into the DistanceSolver framework without it being clunky.

cassiopeia/solver/SharedMutationJoiningSolver.py
richardyz98: What would be the typing for a numba compiled function out of curiosity?

cassiopeia/data/CassiopeiaTree.py
richardyz98: Do you mean we should change all of our functions to work with flattened dissimilarity maps? Or do you mean that there's a way to pass a flattened array into a pandas dataframe. I think the issue is that with flattened arrays it would be difficult to maintain indices (as we do in the next line when we transform the dissimilarity map into a pandas dataframe). I think that if we do not transform it into a df it might hurt usability

mattjones315: Ideally, we should change all of our solver methods to work with flattened dissimilarity matrices. It's just a lot more efficient! But that's not necessarily a top priority right now. Let's just not return, by default, a full dissimilarity matrix -- I'd rather have to convert it to squareform elsewhere.

cassiopeia/solver/SharedMutationJoiningSolver.py
mattjones315: I see your point. I don't see a right way to do this either, so we'll just have to live with duplicated code for now. Thanks for clarifying!

cassiopeia/solver/SharedMutationJoiningSolver.py
mattjones315: I think you can still call it a `Callable`, right? Does this break the `numba` compilation?

cassiopeia/solver/SharedMutationJoiningSolver.py
```diff
@@ -0,0 +1,309 @@
...
+
+        character_matrix.drop(
+            index=[cherry[0], cherry[1]],
+            inplace=True,
+        )
+
+        return similarity_map
+
+    @staticmethod
+    @numba.jit(nopython=True)
```
mattjones315, line 266: Why do we have so many spaces here? Can we clean this up?

mattjones315 (commented):
Looks good to me, Richard! I have one really small formatting change I'm going to make and then I'll merge it in!

mattjones315 (approved):

PR #74, opened 2021-03-01
Add concrete class TumorWithAFitSubclone that implements a TreeSimulator

sprillo:
I also added bulk `get_times` and `set_times` methods to the `CassiopeiaTree`. I still need to add tests in CassiopeiaTree for these two methods, but before let's make sure that you like these methods and their semantics too.

### DIFF
--- a/cassiopeia/data/CassiopeiaTree.py
+++ b/cassiopeia/data/CassiopeiaTree.py
@@ -557,6 +557,40 @@ class CassiopeiaTree:
                 self.get_time(child) - new_time
             )
 
+    def set_times(self, time_dict: Dict[str, float]) -> None:
+        """Sets the time of all nodes in the tree.
+
+        Importantly, this maintains consistency with the rest of the tree. In
+        other words, setting the time of all nodes will change the length of
+        the edges too. This function requires monotonicity of times are
+        maintained (i.e. no negative branch lengths).
+
+        Args:
+            time_dict: Dictionary mapping nodes to their time.
+
+        Raises:
+            CassiopeiaTreeError if the tree is not initialized, or if the time
+            of any parent is greater than that of a child.
+        """
+        self.__check_network_initialized()
+
+        # TODO: Check that the keys of time_dict match exactly the nodes in the
+        # tree and raise otherwise?
+        # Currently, if nodes are missing in time_dict, code below blows up. If
+        # extra nodes are present, they are ignored.
+
+        for (parent, child) in self.edges:
+            time_parent = time_dict[parent]
+            time_child = time_dict[child]
+            if time_parent > time_child:
+                raise CassiopeiaTreeError(
+                    "Time of parent greater than that of child: "
+                    f"{time_parent} > {time_child}"
+                )
+            self.__network[parent][child]["length"] = time_child - time_parent
+        for node, time in time_dict.items():
+            self.__network.nodes[node]["time"] = time
+
     def get_time(self, node: str) -> float:
         """Gets the time of a node.
 
@@ -569,6 +603,19 @@ class CassiopeiaTree:
         self.__check_network_initialized()
 
         return self.__network.nodes[node]["time"]
+
+    def get_times(self) -> Dict[str, float]:
+        """Gets the times of all nodes.
+
+        Returns the times of all nodes, defined as the sum of edge lengths from
+        the root to that node.
+
+        Raises:
+            CassiopeiaTreeError if the tree has not been initialized.
+        """
+        self.__check_network_initialized()
+
+        return dict([(node, self.get_time(node)) for node in self.nodes])
 
     def set_branch_length(self, parent: str, child: str, length: float):
         """Sets the length of a branch.
--- a/cassiopeia/simulator/SimpleFitSubcloneSimulator.py
+++ b/cassiopeia/simulator/SimpleFitSubcloneSimulator.py
@@ -1,20 +1,20 @@
 """
-This file stores a subclass of TreeSolver, the TumorWithAFitSubclone. The
-TumorWithAFitSubclone simulates a tumor phylogeny which develops one fit
-subclone.
+This file stores a subclass of TreeSolver, the SimpleFitSubcloneSimulator. The
+SimpleFitSubcloneSimulator simulates a clonal population which develops one
+fit subclone.
 """
 import networkx as nx
 from queue import Queue
 
 from cassiopeia.data import CassiopeiaTree
-from .TreeSimulator import TreeSimulator
+from cassiopeia.simulator import TreeSimulator
 
 
-class TumorWithAFitSubclone(TreeSimulator):
+class SimpleFitSubcloneSimulator(TreeSimulator):
     r"""
-    Simulates a tumor phylogeny which develops one fit subclone.
+    Simulates a clonal population which develops one fit subclone.
 
-    This TreeSimulator simulates a (binary) tumor phylogeny that evolves
+    This TreeSimulator simulates a (binary) clonal population that evolves
     neutrally with constant branch length 'branch_length_neutral', until in
     generation 'generations_until_fit_subclone' one of the lineages develops
     fitness and starts to expand with the new branch length 'branch_length_fit'.
@@ -22,25 +22,26 @@ class TumorWithAFitSubclone(TreeSimulator):
     This TreeSimulator is useful because (1) it generates deterministic
     phylogenies and (2) the phylogenies are as simple as possible while being
     non-neutral. The ability to control the branch lengths of the neutrally
-    evolving cells and of the fit cells allows us to simulate different degrees
-    of fitness.
+    evolving population and of the fit subpopulation allows us to simulate
+    different degrees of fitness.
 
     Note that the branches for the leaves of the tree need not have length
-    exactly equal to branch_length_neutral nor branch_length_fit, because the
-    cells get assayed at the arbitrary time specified by 'experiment_duration',
-    which is not necessarily the time at which they divide.
+    exactly equal to 'branch_length_neutral' nor 'branch_length_fit', because
+    the population gets assayed at the arbitrary time specified by
+    'experiment_duration', which is not necessarily the time at which a
+    division even happens.
 
     The names of the leaf nodes will be given by a unique integer ID, followed
-    by "_neutral" for the neutrally evolving cells, and by "_fit" for the fit
-    cells.
+    by "_neutral" for the neutrally evolving individuals, and by "_fit" for the
+    fit individuals.
 
     Finally, note that we do not enforce the condition branch_length_neutral >
     branch_length_fit, therefore it is possible to make the subclone less fit,
     i.e. expand slower.
 
     Args:
-        branch_length_neutral: Branch length of the neutrally evolving cells.
-            All cells are neutrally evolving until generation
+        branch_length_neutral: Branch length of the neutrally evolving
+            individuals. All individuals are neutrally evolving until generation
             'generations_until_fit_subclone', when exactly one of the lineages
             develops fitness.
         branch_length_fit: The branch length of the fit subclone, which appears
@@ -101,7 +102,7 @@ class TumorWithAFitSubclone(TreeSimulator):
             )
             time_of_division = time + time_till_division
             if time_of_division >= experiment_duration:
-                # Not enough time left for the cell to divide.
+                # Not enough time left for the individual to divide.
                 times[node] = experiment_duration
                 continue
             # Create children, add edges to them, and push children to the
--- a/cassiopeia/simulator/SimpleFitSubcloneSimulator.py
+++ b/cassiopeia/simulator/SimpleFitSubcloneSimulator.py
@@ -5,9 +5,10 @@ fit subclone.
 """
 import networkx as nx
 from queue import Queue
+from typing import Generator, Tuple, Union
 
 from cassiopeia.data import CassiopeiaTree
-from cassiopeia.simulator import TreeSimulator
+from cassiopeia.simulator.TreeSimulator import TreeSimulator
 
 
 class SimpleFitSubcloneSimulator(TreeSimulator):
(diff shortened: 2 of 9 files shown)


cassiopeia/simulator/TumorWithAFitSubclone.py
```diff
@@ -0,0 +1,141 @@
...
+from cassiopeia.data import CassiopeiaTree
+from .TreeSimulator import TreeSimulator
+
+
+class TumorWithAFitSubclone(TreeSimulator):
+    r"""
+    Simulates a tumor phylogeny which develops one fit subclone.
+
+    This TreeSimulator simulates a (binary) tumor phylogeny that evolves
+    neutrally with constant branch length 'branch_length_neutral', until in
```
mattjones315, line 18: I think we should remove explicit references to tumors, since this is useful in other contexts

cassiopeia/simulator/TumorWithAFitSubclone.py
```diff
@@ -0,0 +1,141 @@
...
+        branch_length_neutral = self.branch_length_neutral
+        branch_length_fit = self.branch_length_fit
+        experiment_duration = self.experiment_duration
+        generations_until_fit_subclone = self.generations_until_fit_subclone
+
+        def node_name_generator():
+            i = 0
+            while True:
+                yield str(i)
+                i += 1
```
mattjones315, line 78: This is pretty cool -- we should use this elsewhere in our tree solvers!

cassiopeia/simulator/TumorWithAFitSubclone.py
```diff
@@ -0,0 +1,141 @@
...
+        times = {}
+
+        root = next(names) + "_neutral"
+        tree.add_node(root)
+        times[root] = 0.0
+
+        root_child = next(names) + "_neutral"
+        tree.add_edge(root, root_child)
+        q.put((root_child, 0.0, "neutral", 0))
+        subclone_started = False
```
mattjones315, line 93: Why do you add a single child here? I would have imagined adding two children from the root. Why not just add the root directly to the queue?

mattjones315 (commented):
Hey Sebastian -- this simple simulator looks great! I've left some comments/questions that I'd like to get resolved before we merge it in.

But some main takeaways -- 

(1) I think we should move away from the name `TumorWithAFitSubclone`, and even make it more clear that this is a _very_ simple interpretation of "fitness". My suggestion is to call it `SimpleFitSubcloneSimulator` which gets rid of the reference to tumor biology and makes it clear that it's simple.
(2) On the topic of simplicity, I understand why it's nice to have a deterministic simulator and I don't want to tell you what is and is not useful for your work. However, I don't know how generally useful this would be -- in my mind, simulators should have some randomness built into them so you can get an idea of a what a distribution of test cases or examples might look like. One way we can build randomness into this is to specify a distribution of waiting times for the fit subclone rather than an explicit length. We could actually support both by taking in a function (HELLO COMPOSITIONAL DESIGN!) for the waiting times. 
(3) The last item that I'd like to bring up is how we design functions in the `CassiopeiaTree` class. I think it's best to rely on the API, even internally (e.g. `set_times` should use `set_time`). Is there a reason not to do this?

Thanks in advance for addressing these comments / questions!

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -557,6 +557,40 @@ def set_time(self, node: str, new_time: float) -> None:
...
+            time_child = time_dict[child]
+            if time_parent > time_child:
+                raise CassiopeiaTreeError(
+                    "Time of parent greater than that of child: "
+                    f"{time_parent} > {time_child}"
+                )
+            self.__network[parent][child]["length"] = time_child - time_parent
+        for node, time in time_dict.items():
+            self.__network.nodes[node]["time"] = time
+
```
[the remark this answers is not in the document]
sprillo: > I think if you were to traverse the tree correctly while updating, you should avoid any exceptions being raised.

This is exactly why I decided not to use `set_time`: we would have to do some juggling to avoid exceptions. We can think about this though, I agree having the logic in only one place would be ideal. Another hacky solution: we can have a `pedantic` boolean flag in the `set_time` method, which by default is `True`, but when set to `False` will not enforce any constraints. It can be hidden in a private method such that it is not exposed to users.

cassiopeia/simulator/TumorWithAFitSubclone.py
sprillo: Because the founding cell doesn't divide immediately.

sprillo: (In particular the resulting tree is not binary)

sprillo: Sorry if I misunderstood the question BTW, maybe we should chat about it.

sprillo:
@mattjones315 regarding point #2:

> (2) On the topic of simplicity, I understand why it's nice to have a deterministic simulator and I don't want to tell you what is and is not useful for your work. However, I don't know how generally useful this would be -- in my mind, simulators should have some randomness built into them so you can get an idea of a what a distribution of test cases or examples might look like. One way we can build randomness into this is to specify a distribution of waiting times for the fit subclone rather than an explicit length. We could actually support both by taking in a function (HELLO COMPOSITIONAL DESIGN!) for the waiting times.

I really like the idea of using composition to inject the waiting times and get different behaviors. My concern is: how would you decide when the fit subclone starts expanding? Right now I just specify the generation number; since all branch lengths are the same, there is a very clear notion of "generation 6". What fo you propose in general? If we find a simple way to decide when the subclone starts to expand, I would be happy implementing the more general version you propose.

On the topic of randomness, I think it can be useful later on to perform more thorough quantitative analysis. However, I think any method development should _start_ with a deterministic playground that is as simple as possible, and later on move to more complicated settings. I think this TreeSimulator is as simple as possible!

cassiopeia/simulator/TumorWithAFitSubclone.py
```diff
@@ -0,0 +1,141 @@
+"""
+This file stores a subclass of TreeSolver, the TumorWithAFitSubclone. The
+TumorWithAFitSubclone simulates a tumor phylogeny which develops one fit
+subclone.
+"""
+import networkx as nx
+from queue import Queue
+
+from cassiopeia.data import CassiopeiaTree
+from .TreeSimulator import TreeSimulator
```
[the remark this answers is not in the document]
sprillo: Good point. I don't know what the difference between the two ways is from a technical point of view, nor from a design perspective. I'm fine changing this to `from cassiopeia.simulator import TreeSimulator` for consistency.

cassiopeia/data/CassiopeiaTree.py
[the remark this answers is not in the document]
mattjones315: Ah, I sort of like the `pedantic` solution. 

Another solution -- you could do a bottom-up pass to begin with setting all times to be `np.inf`, and then do a top-down pass setting the times to those in the `time_dict`. I think this is guaranteed to not break anything, right?

cassiopeia/simulator/TumorWithAFitSubclone.py
mattjones315: You're totally right - the founding cell doesn't divide immediately, but it also doesn't affect the topology at all.

The time that it takes for a cell to divide can be encoded in the branch lengths leading to both of its children, right? In general, I think the simulations should return a tree that is well-formed, in the sense that each node should have at least 2 children (unless it's a leaf). Does that sound reasonable?

mattjones315:
> @mattjones315 regarding point #2:
> 
> > (2) On the topic of simplicity, I understand why it's nice to have a deterministic simulator and I don't want to tell you what is and is not useful for your work. However, I don't know how generally useful this would be -- in my mind, simulators should have some randomness built into them so you can get an idea of a what a distribution of test cases or examples might look like. One way we can build randomness into this is to specify a distribution of waiting times for the fit subclone rather than an explicit length. We could actually support both by taking in a function (HELLO COMPOSITIONAL DESIGN!) for the waiting times.
> 
> I really like the idea of using composition to inject the waiting times and get different behaviors. My concern is: how would you decide when the fit subclone starts expanding? Right now I just specify the generation number; since all branch lengths are the same, there is a very clear notion of "generation 6". What fo you propose in general? If we find a simple way to decide when the subclone starts to expand, I would be happy implementing the more general version you propose.
> 
> On the topic of randomness, I think it can be useful later on to perform more thorough quantitative analysis. However, I think any method development should _start_ with a deterministic playground that is as simple as possible, and later on move to more complicated settings. I think this TreeSimulator is as simple as possible!

Ah, these are interesting points! Definitely, it's nice to have a deterministic framework to have you understand the problem. But when we do method development we like to have some noise in our examples so we can get a feel for the robustness of the method. 

To your point -- I think you can maintain a generation counter and still initiate the fitness at the `nth` generation. But it'd be nice to have some randomness in the edge lengths (or at least support randomness).

cassiopeia/simulator/TumorWithAFitSubclone.py
sprillo: I love these subtleties here! Let me try to clarify what's on my mind. My interpretation of a single-cell phylogeny is as follows:
- The root of the tree is the "founding cell" event.
- An internal node of the phylogeny which is not root, corresponds to a cell division event.
- Each branch of the phylogeny corresponds to the lifetime of a cell.
With this interpretation:
- The branch lengths leading from the root to each of its children are the times it takes for each of the daughter cells to divide - which is distinct from the time it takes for the founding cell to divide.
- The time it takes for the founding cell to divide is encoded in the first branch of the tree.

Regarding the comment that `each node should have at least 2 children`, I agree with this for internal nodes different from the root, but the way I interpret the phylogeny, the root should have degree 1, and in my mind this reflects the biological process more accurately. Here's one question to put my interpretation into perspective: where on the phylogeny should the CRISPR/Cas9 mutations that hit the founding cell - before it divides - be mapped to, if the root is required to have two children?

I'm happy to talk about this in more depth!

sprillo:
> > @mattjones315 regarding point #2:
> > > (2) On the topic of simplicity, I understand why it's nice to have a deterministic simulator and I don't want to tell you what is and is not useful for your work. However, I don't know how generally useful this would be -- in my mind, simulators should have some randomness built into them so you can get an idea of a what a distribution of test cases or examples might look like. One way we can build randomness into this is to specify a distribution of waiting times for the fit subclone rather than an explicit length. We could actually support both by taking in a function (HELLO COMPOSITIONAL DESIGN!) for the waiting times.
> > 
> > 
> > I really like the idea of using composition to inject the waiting times and get different behaviors. My concern is: how would you decide when the fit subclone starts expanding? Right now I just specify the generation number; since all branch lengths are the same, there is a very clear notion of "generation 6". What fo you propose in general? If we find a simple way to decide when the subclone starts to expand, I would be happy implementing the more general version you propose.
> > On the topic of randomness, I think it can be useful later on to perform more thorough quantitative analysis. However, I think any method development should _start_ with a deterministic playground that is as simple as possible, and later on move to more complicated settings. I think this TreeSimulator is as simple as possible!
> 
> Ah, these are interesting points! Definitely, it's nice to have a deterministic framework to have you understand the problem. But when we do method development we like to have some noise in our examples so we can get a feel for the robustness of the method.
> 
> To your point -- I think you can maintain a generation counter and still initiate the fitness at the `nth` generation. Even though branch lengths are not uniform, you can still count generations as the # of cell divisions up to a certain node. I think it'd be helpful to have some randomness in the edge lengths (or at least support randomness).

Sounds good Matt! I will think about the best way to implement this.

cassiopeia/simulator/TumorWithAFitSubclone.py
mattjones315: For posterity: we decided that having a founder cell with a single child makes more sense from a modeling perspective. It doesn't change the tree topology, but will have several downstream ramifications when overlaying the lineage tracing data and inferring branch lengths. For this reason we're keeping it here.

cassiopeia/simulator/SimpleFitSubcloneSimulator.py
```diff
@@ -1,67 +1,91 @@
...
         generations_until_fit_subclone: The generation at which one lineage
             develops fitness, giving rise to a fit subclone.
     """
 
     def __init__(
         self,
-        branch_length_neutral: float,
-        branch_length_fit: float,
+        branch_length_neutral: Union[float, Generator[float, None, None]],
+        branch_length_fit: Union[float, Generator[float, None, None]],
```
mattjones315, line 62: I'm a bit confused why we have to use `Generators` here? The way I had imagined doing this is by passing some function for returning a branch lengths.

For example -- in the deterministic case: 

```
branch_length_neutral = lambda: 1
branch_length_neutral = lambda: 0.5
```

And in the case where you'd like to draw from some distribution:

```
branch_length_neutral = lambda: np.random.exponential(1)
branch_length_neutral = lambda: np.random.exponential(0.5)
```

Why is a `Generator` preferred to this? Is this because you'd like to support users just passing in a float rather than an anonymous function that always returns a constant?

test/simulator_tests/simple_fit_subclone_simulator_test.py
```diff
@@ -0,0 +1,76 @@
...
+        def branch_length_fit_generator() -> Generator[float, None, None]:
+            while True:
+                yield np.random.exponential(0.5)
+
+        tree = SimpleFitSubcloneSimulator(
+            branch_length_neutral=branch_length_neutral_generator(),
+            branch_length_fit=branch_length_fit_generator(),
+            experiment_duration=4.9,
+            generations_until_fit_subclone=2,
+        ).simulate_tree()
```
mattjones315, line 67: I think it's a bit odd to expect users to pass in these generators. I understand you want to support both deterministic and random branch lengths, which I agree is nice, but I think there's a more straightforward way to have branch lengths be drawn from a random distribution. Could you have `SimpleFitSubcloneSimulator` take in a function or a constant and turn either of these into an anonymous function or a Generator?  Or have `SimpleFitSubcloneSimulator` take in an anonymous function altogether?

mattjones315 (commented):
Hey Sebastian, this is awesome!

I think the use of a Generator is a bit odd (I understand it's super pythonic, but I'm a bit worried that it'd be too complicated for a simple user). Can we consider alternative options for incorporating random functions into the branch length simulator? I left a few suggestions, perhaps these are reasonable?

cassiopeia/simulator/SimpleFitSubcloneSimulator.py
sprillo: You're totally right Matt, we don't need to use a generator at all! We can use a callable. I'll make your proposed change, I agree it's much simpler.

mattjones315 (approved):
This looks great! I think it's ready to be merged in.

PR #75, opened 2021-03-05
[Ready for Review] General bd simulator

sprillo:


cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,336 @@
...
+                        )[0]
+                        tree.edges[parent, remaining_lineage["id"]][
+                            "weight"
+                        ] += (min_total_time - remaining_lineage["total_time"])
+                        tree.nodes[remaining_lineage["id"]][
+                            "birth_scale"
+                        ] = tree.nodes[parent]["birth_scale"]
+                    break
+            # If extant tips are the stopping criteria, pop the minimum age
+            # lineage at each step
```
sprillo, line 219: Popping the minimum age lineage works regardless of whether the stopping condition is num_extant or experiment_time.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,336 @@
...
+                    [i["total_time"] for i in current_lineages]
+                )
+                lineage = current_lineages.pop(min_age_ind)
+            else:
+                lineage = current_lineages.pop(0)
+            for _ in range(2):
+                unique_id = sample_event(unique_id, lineage)
+
+        # Prune dead lineages and collapse resulting unifurcations
+        if death_waiting_dist and len(tree.nodes) > 1:
```
sprillo, line 231: it shouldn't matter whether or not `death_waiting_dist` is provided, right? If it is not provided, the code below simply won't have anything to prune, but should still run?

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,336 @@
...
+                    if (
+                        tree.out_degree(i) == 0
+                        and tree.nodes[i]["total_time"] < experiment_time
+                    ):
+                        self.remove_and_prune_lineage(i, tree)
+            if num_extant:
+                surviving_ids = [i["id"] for i in current_lineages]
+                for i in list(tree.nodes):
+                    if tree.out_degree(i) == 0 and i not in surviving_ids:
+                        self.remove_and_prune_lineage(i, tree)
```
sprillo, line 243: For testing purposes it might be really cool to have an option to *not* prune the dead lineages, so that we can see where they are and that they look correct. I think an argument `prune_dead_lineages` in `__init__` would be ideal? However, this might not be easy to get working because I think `CassiopeiaTree` assumes that all leaves in the tree are extant...

One thing we *can* easily do though is to allow not collapsing unifurcations. That way, at least we get to see where in the tree the lineages that got extinct branched off. Again, this is only for testing / visualization purposes.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,336 @@
...
+
+        # If only implicit root remains after pruning dead lineages, error
+        if len(tree.nodes) == 1:
+            raise BirthDeathFitnessError(
+                "All lineages died before stopping condition"
+            )
+
+        return CassiopeiaTree(tree=tree)
+
+    def remove_and_prune_lineage(self, node: int, tree: nx.DiGraph):
```
sprillo, line 255: This method sounds more generally useful? Perhaps we can put it into `cassiopeia/data/utilities.py` or something?

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,336 @@
...
+                tree.out_degree(curr_parent) < 1
+                and tree.in_degree(curr_parent) > 0
+            ):
+                next_parent = list(tree.predecessors(curr_parent))[0]
+                tree.remove_node(curr_parent)
+                curr_parent = next_parent
+
+    def collapse_unifurcations(
+        self, tree: nx.DiGraph, source: int = None
+    ) -> None:
```
sprillo, line 281: Same comment as above: Perhaps we can put it into `cassiopeia/data/utilities.py` or something?

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,336 @@
...
+    birth_waiting_dist,
+    birth_scale_param,
+    death_waiting_dist,
+    fitness_num_dist = fitness_num_dist,
+    fitness_strength_dist=fitness_strength_dist,
+    num_extant=8,
+#     experiment_time = 1
+)
+
+"""
```
sprillo, line 336: I think we can put this directly in the docstring of the class.

test/simulator_tests/birth_death_simulator_test.py
```diff
@@ -0,0 +1,530 @@
...
+from cassiopeia.simulator.BirthDeathFitnessSimulator import (
+    BirthDeathFitnessError,
+)
+
+
+def get_leaves(tree):
+    return [n for n in tree.nodes if tree.out_degree(n) == 0]
+
+
+def test_tree(tree):
```
sprillo, line 18: Can we add type hints and a docstring for what this function is doing?

test/simulator_tests/birth_death_simulator_test.py
```diff
@@ -0,0 +1,530 @@
...
+
+        tree_top = tree.get_tree_topology()
+        results = test_tree(tree_top)
+        for i in results[0]:
+            self.assertTrue(np.isclose(i, 2))
+        self.assertTrue(results[2])
+        self.assertNotIn(10, tree_top.nodes)
+        self.assertNotIn(2, tree_top.nodes)
+
+    def test_nonconstant_yule_with_predictable_fitness(self):
```
sprillo, line 397: Can we add a docstring (even if small / one liner) to the tests just to understand at a high level what is being tested in each function?

sprillo (commented):
This is looking amazing! Just have some questions around documentation, and about understanding what exactly the main loop of the algorithm is doing (`sample_event`). Once that is more clearly documented such that I have a better understanding, I'll re-read the code to verify that the logic is correct.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
mattjones315: I agree. This should actually already be implemented in `cassiopeia.data.utilities`. 

At any rate, `ete3.Tree.prune` has very similar functionality. It'd be great to use one of these tools here.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
mattjones315: Yes, I believe this already exists in the `data.utiliites` module, but perhaps only for ete3 trees. Let's add this function there and call it from that library.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,409 @@
...
+from cassiopeia.data.CassiopeiaTree import CassiopeiaTree
+from cassiopeia.simulator import TreeSimulator
+
+
+class BirthDeathFitnessError(Exception):
+    """An Exception class for all exceptions generated by the
+    GeneralBirthDeathSimulator
+    """
+
+    pass
```
mattjones315, line 22: I think it would be better to have a `TreeSimulatorError` class that we can use for these exceptions

test/simulator_tests/birth_death_simulator_test.py
```diff
@@ -0,0 +1,573 @@
...
+    return times, len(leaves), correct_degrees
+
+
+class BirthDeathSimulatorTest(unittest.TestCase):
+    def test_remove_and_prune_lineage_all(self):
+        bd_sim = BirthDeathFitnessSimulator()
+        tree = nx.balanced_tree(2, 3, create_using=nx.DiGraph())
+        leaves = get_leaves(tree)
+        for i in leaves:
+            bd_sim.remove_and_prune_lineage(i, tree)
```
mattjones315, line 52: This is a good example for why we'd like to have this function in the `data.utilities` module -- it's super useful and doesn't make sense to have to create a `bd_sim` object to use it

test/simulator_tests/birth_death_simulator_test.py
```diff
@@ -0,0 +1,573 @@
...
+            (0, 2, {"weight": 1.5}),
+            (2, 9, {"weight": 1.5}),
+            (2, 4, {"weight": 3.0}),
+            (2, 6, {"weight": 3.0}),
+            (6, 7, {"weight": 1.5}),
+            (6, 8, {"weight": 1.5}),
+        ]
+        self.assertEqual(list(tree.edges(data=True)), expected_edges)
+
+    def test_collapse_unifurcations_2(self):
```
mattjones315, line 143: Can we be more descriptive with our names? What exactly is this test doing?

mattjones315 (commented):
Hey Richard/Sebastian,

This is coming along great! I left some comments that I hope will be useful, though I know this is still WIP. Let me know if there's anything else I can help with!

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -34,18 +35,21 @@ def simulate_tree(
     ) -> CassiopeiaTree:
```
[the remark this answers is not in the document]
sprillo: Nice catch @mattjones315 ! I totally missed this. Yes, all the hyperparameters should go into `__init__`

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,409 @@
...
+                    if num_mutations < 0:
+                        raise BirthDeathFitnessError(
+                            "Negative number of mutations detected"
+                        )
+                    for _ in range(num_mutations):
+                        total_birth_mutation_strength *= fitness_strength_dist()
+                        if total_birth_mutation_strength < 0:
+                            raise BirthDeathFitnessError(
+                                "Negative mutation strength detected"
+                            )
```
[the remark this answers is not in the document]
sprillo: Yeah, you are right: The current implementation assumes positive multiplicative fitness factors. But we can accept negative fitnesses and use the `exp` function to map them to positive multiplicative factors. The current implementation is more general, but if this level of generality does not seem useful to the general user, we can change it. Of course, the documentation of the simulator should be 100% clear what is going on (the `exp` mapping for example, if we were to use it).

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,409 @@
...
+                            + lineage["total_time"],
+                            "active": False,
+                        },
+                    )
+                )
+                return unique_id + 1
+
+        # Instantiate the implicit root
+        tree = nx.DiGraph()
+        tree.add_node(0)
```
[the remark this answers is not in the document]
sprillo: Also, nodes **should** be strings (although currently CassiopeiaTree does not enforce it)

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,409 @@
...
+                    for lineage in remaining_lineages:
+                        parent = list(tree.predecessors(lineage["id"]))[0]
+                        tree.edges[parent, lineage["id"]]["weight"] += (
+                            min_total_time - lineage["total_time"]
+                        )
+                        tree.nodes[lineage["id"]]["birth_scale"] = tree.nodes[
+                            parent
+                        ]["birth_scale"]
+                        observed_nodes.append(lineage["id"])
+                    break
```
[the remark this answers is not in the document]
sprillo: @richardyz98 one proposal: just make all current events inactive, as I suggested in our chat.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,409 @@
...
+            # Pop the minimum age lineage to simulate forward time
+            _, _, lineage = current_lineages.get()
+            for _ in range(2):
+                unique_id = sample_lineage_event(unique_id, lineage)
```
sprillo, line 333: @richardyz98 I would rather we only call `sample_lineage_event(unique_id, lineage)` if the lineage is active. If it is inactive, the function `sample_lineage_event(unique_id, lineage)` does nothing anyway, so the end result is the same. Thus, we might as well put the check here. In `sample_lineage_event(unique_id, lineage)` we can further assert that the lineage is active. This will make the code a lot easier to understand because the invariants will be clearer.

test/simulator_tests/birth_death_simulator_test.py
sprillo: Yeah, generally, I think we should document all our tests.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,361 @@
...
+        self.fitness_distribution = fitness_distribution
+        self.num_extant = num_extant
+        self.experiment_time = experiment_time
+
+    def simulate_tree(
+        self,
+    ) -> CassiopeiaTree:
+        """Simulates trees from a general birth/death process with fitness.
+
+        A forward-time birth/death process is simulated by maintaining a queue
```
sprillo, line 65: One comment I have about the documentation is that we should be able to describe the generative model without revealing any implementation details at all. Note that you are revealing implementation details by saying `maintaining a queue`, `At each live currently extant node`, etc. Instead, we can simply say that:
- All cells have the same death waiting time, given by `death_waiting_distribution`.
- Each cell has a specific birth scale parameter, where:
  - The root has birth scale parameter `initial_birth_scale`
  - Whenever a cell divides, it's birth rate scale parameter changes as follows: [...]

And that's the end of if. We don't have to say anything else. All else is an implementation detail, the user doesn't and shouldn't care about it.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,361 @@
...
+        then updated by the total multiplicative strength factor across all
+        mutations and passed on to the descendant nodes.
+
+        There are two stopping conditions for the simulation. The first is
+        "number of extant nodes", which specifies the simulation to stop the
+        first moment the specified number of extant nodes exist. The second is
+        "experiment time", which specifies the time at which to end the
+        experiment, i.e. the experiment ends when all living lineages reach
+        this time in their total lived time. At least one of these two stopping
+        criteria must be provided.
```
sprillo, line 97: (And we will also allow both to be provided.)

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,361 @@
...
+                    added to the tree
+                lineage: The current extant lineage to extend. Contains the ID
+                    of the internal node to attach the descendant to, the
+                    current birth scale parameter of the lineage, the current
+                    total lived time of the lineage, and the status of whether
+                    the lineage is still dividing
+
+            Returns:
+                The new unique ID to use for the next event, incremented only if
+                a new descendant is added to the tree
```
sprillo, line 175: We should update the docstring: this function no longer return anything.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,361 @@
...
+                    weight=self.experiment_time - lineage["total_time"],
+                )
+                tree.nodes[unique_id]["total_time"] = self.experiment_time
+
+                current_lineages.put(
+                    (
+                        self.experiment_time,
+                        unique_id,
+                        {
+                            "id": unique_id,
```
sprillo, line 211: Why do we have the `unique_id` twice in this data structure?

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,361 @@
...
+                        if num_mutations < 0:
+                            raise TreeSimulatorError(
+                                "Negative number of mutations detected"
+                            )
+                        for _ in range(num_mutations):
+                            base_selection_coefficient *= self.fitness_distribution()
+                            if base_selection_coefficient < 0:
+                                raise TreeSimulatorError(
+                                    "Negative mutation strength detected"
+                                )
```
sprillo, line 235: Let's move the logic of computing the new birth scale parameter to a separate function; it will make the code more readable and easier to refactor if we want to change the way that birth scales are updated.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,361 @@
...
+                            "total_time": self.experiment_time,
+                            "active": False,
+                        },
+                    )
+                )
+                observed_nodes.append(unique_id)
+
+            else:
+                if birth_waiting_time < death_waiting_time:
+                    # Update fitness
```
sprillo, line 222: Why don't we just say "update birth rate"?

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,361 @@
...
+                    for lineage in remaining_lineages:
+                        parent = list(tree.predecessors(lineage["id"]))[0]
+                        tree.edges[parent, lineage["id"]]["weight"] += (
+                            min_total_time - lineage["total_time"]
+                        )
+                        tree.nodes[lineage["id"]]["birth_scale"] = tree.nodes[
+                            parent
+                        ]["birth_scale"]
+                        observed_nodes.append(lineage["id"])
+            if not current_lineages.empty():
```
sprillo, line 336: I'm going to disagree with @mattjones315  here: I think we should `break` in the above `if self.num_extant:` statement, and remove this `if not current_lineages.empty():` statement altogether, since I find really weird given that we already made the check `if not current_lineages.empty():` in the outer loop.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,361 @@
...
+                    for _ in range(2):
+                        sample_lineage_event(lineage)
+
+        # Prune dead lineages and collapse resulting unifurcations
+        if len(tree.nodes) > 1:
+            for i in list(tree.nodes):
+                if tree.out_degree(i) == 0 and i not in observed_nodes:
+                    utilities.remove_and_prune_lineage(i, tree)
+            if len(tree.nodes) > 1:
+                utilities.collapse_unifurcations(tree, source="1")
```
sprillo, line 351: Let's have an argument `collapse_unifurcations: bool = True` in the class' `__init__` method, that toggles this logic. I would like to be able to get trees _with_ the unifurcations in them too.

test/simulator_tests/birth_death_simulator_test.py
```diff
@@ -0,0 +1,334 @@
...
+    BirthDeathFitnessSimulator,
+)
+from cassiopeia.simulator.TreeSimulator import (
+    TreeSimulatorError,
+)
+
+import cassiopeia.data.utilities as utilities
+
+
+def get_leaves(tree):
```
sprillo, line 16: missing type hints.

sprillo (commented):
Almost there! Left a couple more comments.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
richardyz98: This is to provide an additional field so that the PriorityQueue can break ties in total_time. Otherwise, if two lineages have the same total time, then it is unable to break ties. I also wanted to keep the total_time and unique_id information in the lineage dictionary itself too, for readability purposes. Let me know if you disagree!

sprillo: Ah, I see, this is a very good reason, I hadn't noticed. Thanks Richard!

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,399 @@
...
+                                "id": unique_id,
+                                "birth_scale": lineage["birth_scale"],
+                                "total_time": death_waiting_time
+                                + lineage["total_time"],
+                                "active": False,
+                            },
+                        )
+                    )
+
+        def update_birth_scale(birth_scale):
```
sprillo, line 286: Shouldn't this function take in `self`?

richardyz98: No I don't think so, as it's defined in the body of `simulate tree`

sprillo: Oh I didn't notice! Thanks Richard!

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,399 @@
...
+        death_waiting_distribution: Optional[
+            Callable[[], float]
+        ] = lambda: np.inf,
+        mutation_distribution: Optional[Callable[[], int]] = None,
+        fitness_distribution: Optional[Callable[[], float]] = None,
+        fitness_base: float = np.e,
+        num_extant: Optional[int] = None,
+        experiment_time: Optional[float] = None,
+        collapse_unifurcations: bool = True,
+    ):
```
mattjones315, line 112: I propose we add the option to add a `random_state` or `random_seed` (preferably a state) for determinism. This will be nice when distributing our code / simulations as well as for testing purposes below.

test/simulator_tests/birth_death_simulator_test.py
```diff
@@ -0,0 +1,419 @@
...
+
+    out_degrees = [tree.out_degree(n) for n in tree.nodes]
+    out_degrees.pop(0)
+
+    correct_degrees = all(x == 2 or x == 0 for x in out_degrees)
+
+    return times, len(leaves), correct_degrees
+
+
+class BirthDeathSimulatorTest(unittest.TestCase):
```
mattjones315, line 50: I think you can make a lot of the simulations below more deterministic with the addition of a numpy random state to the birth death simulator.

mattjones315 (commented):
This looks awesome! It's well documented, well tested, and well written. I think it's nearly there. I did leave some comments that I think will be helpful to resolve for future uses. They're mainly around the right usage of the new utilities added (pruning lineages & collapsing unifurcations) as well as some best practices in functions (like allowing the usage of a numpy random state). Hope this helps and I'm looking forward to merging this in soon!

cassiopeia/data/utilities.py
```diff
@@ -119,6 +119,77 @@ def _to_newick_str(g, node):
...
+            t = tree.get_edge_data(source, succs[0])["weight"]
+            for i in tree.successors(succs[0]):
+                t_ = tree.get_edge_data(succs[0], i)["weight"]
+                tree.add_edge(source, i, weight=t + t_)
+        else:
+            for i in tree.successors(succs[0]):
+                tree.add_edge(source, i)
+        tree.remove_node(succs[0])
+
+
```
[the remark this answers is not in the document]
sprillo: One remark is that if we perform this operation in the CassiopeiaTree class, we will have to be careful about updating the branch lengths. I think it could be useful to have the collapse unifurcations functionality in a CassiopeiaTree too, but I think a version for networkx is also useful to have. To not duplicate code, I would imagine the CassiopeiaTree implementation leveraging the networkX implementation.

I lean towards accepting the current code and leaving the CassiopeiaTree version as a TODO, just because I think other things might be of higher priority.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,399 @@
...
+            Args:
+                unique_id: The unique ID number to be used to name a new node
+                    added to the tree
+                lineage: The current extant lineage to extend. Contains the ID
+                    of the internal node to attach the descendant to, the
+                    current birth scale parameter of the lineage, the current
+                    total lived time of the lineage, and the status of whether
+                    the lineage is still dividing
+            """
+            assert lineage["active"]
```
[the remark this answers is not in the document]
sprillo: Good idea, we can say something like "function's precondition violated".

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,399 @@
...
+                            "active": False,
+                        },
+                    )
+                )
+                observed_nodes.append(unique_id)
+
+            else:
+                if birth_waiting_time < death_waiting_time:
+                    # Update birth rate
+                    updated_birth_scale = update_birth_scale(
```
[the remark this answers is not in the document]
sprillo: Aren't fitness and birth scale the same (for this model)?

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,399 @@
...
+                for _ in range(2):
+                    sample_lineage_event(lineage)
+
+        # Prune dead lineages and collapse resulting unifurcations
+        if len(tree.nodes) > 1:
+            for i in list(tree.nodes):
+                if tree.out_degree(i) == 0 and i not in observed_nodes:
+                    utilities.remove_and_prune_lineage(i, tree)
+            if self.collapse_unifurcations and len(tree.nodes) > 1:
+                utilities.collapse_unifurcations(tree, source="1")
```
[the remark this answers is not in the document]
sprillo: Same remark as above: implementing this in CassiopeiaTree requires some extra care (and time), so I am inclined to defer for later.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
sprillo: The randomness is abstracted out into `birth_waiting_distribution, death_waiting_distribution, mutation_distribution, fitness_distribution` which are provided by the user, I don't think we need to deal with randomness at all in our code?

richardyz98: Hi Matt, 
Currently I'm using random seeds. I think utilizing a random_state might be hard, because it has a set of built-in distributions that it samples from, but we want the BirthDeathFitnessSimulator to be able to use any given distribution. But you're right let me allow the user to pass in a seed as an input

test/simulator_tests/birth_death_simulator_test.py
richardyz98: Hi Matt,
I think some of these tests should work regardless of stochasticity, and so I want to ensure that. So therefore I've left the seed out on these, while others that require a seed I've included. Do you think this is alright?

test/simulator_tests/birth_death_simulator_test.py
```diff
@@ -0,0 +1,419 @@
...
+    times = []
+    for i in leaves:
+        times.append(tree.nodes[i]["total_time"])
+
+    out_degrees = [tree.out_degree(n) for n in tree.nodes]
+    out_degrees.pop(0)
+
+    correct_degrees = all(x == 2 or x == 0 for x in out_degrees)
+
+    return times, len(leaves), correct_degrees
```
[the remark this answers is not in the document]
richardyz98: I think mostly I just wanted to access the out_degree. Maybe we can make a get_out_degree and get_in_degree method to CassiopeiaTree and just test on the CassiopeiaTree?

cassiopeia/simulator/BirthDeathFitnessSimulator.py
richardyz98: also Sebastian in regards to your comment, I think we can make deterministic distributions to pass in, but let's say we're using a function that defines the exponential with some scale. We'd still encounter randomness in the waiting times we draw from the exponential distribution.

sprillo: I'm not sure I understand, here's how I would create my `birth_waiting_distribution` as a user:
```
random_state = np.random.RandomState(1)
def birth_waiting_distribution():
    random_state.exponential()
```
BirthDeathFitnessSimulator does not need to handle randomness in this case, it will be deterministic. It's up to the user to make `birth_waiting_distribution`, etc. deterministic.

test/simulator_tests/birth_death_simulator_test.py
sprillo: I think all our tests should be deterministic. If not, when a test fails, we might not be able to reproduce it.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
richardyz98: Ah I see Sebastian. I think that it may take some of the burden off the user if we allow for a seed though, that way they don't have to create a randomstate for each function. And it also means that they don't need to know about the use/existence of randomstates

test/simulator_tests/birth_death_simulator_test.py
richardyz98: This sounds good I'll add some more seeds

test/simulator_tests/birth_death_simulator_test.py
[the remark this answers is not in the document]
mattjones315: The out degree should just be the number of children that a particular node has, right? The in degree will always be 1.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
mattjones315: To wrap up this conversation from earlier -- I agree with Sebastian that we should rely on users to pass in random states. However, I believe we all agreed that it would be nice to support the use of random seeds for reproducibility.

sprillo: Agreed.

test/simulator_tests/birth_death_simulator_test.py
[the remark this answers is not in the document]
richardyz98: Oh yeah the in_degree will always be 1 except for root, but I wanted to make sure there are no nodes with out_degree 1 (indicating collapse_unifurcations didn't do its job)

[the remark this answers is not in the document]
mattjones315: But you can check this too by looking at the number of children. If `tree.children(n)` == 1 you have a malformed tree and collapse unifurcations didn't do its job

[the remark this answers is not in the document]
richardyz98: that's true. Let me alter the tests

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -863,6 +865,77 @@ def relabel_nodes(self, relabel_map: Dict[str, str]) -> None:
...
+
+            # reset cache because we've changed the tree topology
+            self.__cache = {}
+
+    def collapse_unifurcations(self, source: Optional[int] = None) -> None:
+        """Collapses unifurcations in a given tree.
+
+        Args:
+            tree: The tree to collapse unifurcations on
+            source: The node at which to begin the tree traversal
```
mattjones315, line 901: This documentation is out of date

cassiopeia/simulator/BirthDeathFitnessSimulator.py
```diff
@@ -0,0 +1,404 @@
...
+            A CassiopeiaTree with the tree topology initialized with the
+            simulated tree
+        """
+
+        # Samples whether birth, death, or the end of the experiment comes next
+        # for a given lineage, and any fitness changes
+        def sample_lineage_event(lineage: Dict[str, Union[int, float]]) -> None:
+            """A helper function that samples an event for a lineage.
+
+            Takes a lineage and determines the next event in that lineage's
```
mattjones315, line 176: I have a general design question -- what's the motivation for making these nested functions vs class methods? I think using fewer nested functions makes code more readable, but I'm open to other designs if there's a good reason for it.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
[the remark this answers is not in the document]
mattjones315: I'd like to get rid of this `break` statement before merging in.

cassiopeia/solver/__init__.py
```diff
@@ -9,7 +9,7 @@
 from .MaxCutGreedySolver import MaxCutGreedySolver
 from .MaxCutSolver import MaxCutSolver
 from .NeighborJoiningSolver import NeighborJoiningSolver
-from .solver_utilities import collapse_tree, collapse_unifurcations
+from .solver_utilities import collapse_tree
```
mattjones315, line 12: Would it perhaps be more reasonable to add `collapse_tree` to the `CassiopeiaTree` class as well?

mattjones315 (commented):
Hey Richard, this is looking great!

I don't think you've hit all my comments from before -- specifically, there's an awkward `break` statement I'd like to get rid of. I'm also curious if there's a strong reason for using the nested function design in the simulate function. Can you address my comments before we merge in?

Thanks!

cassiopeia/solver/__init__.py
richardyz98: I was actually thinking of this! But I'd like to save this for a future PR perhaps.

cassiopeia/simulator/BirthDeathFitnessSimulator.py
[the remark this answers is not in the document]
richardyz98: I actually tried a different solution but Sebastian mentioned that he preferred the break. Let me try to continue to think about an elegant solution

cassiopeia/simulator/BirthDeathFitnessSimulator.py
richardyz98: I think I wanted to avoid having to pass around the tree, queue, and name generator to the helper functions. I think that this might be a small price to pay though if we feel it would improve readability. What do you think?

cassiopeia/solver/__init__.py
mattjones315: Sounds good! Can you make a note in the `CassiopeiaTree` class?

cassiopeia/simulator/BirthDeathFitnessSimulator.py
mattjones315: Hmm. Perhaps I'm misreading something, but aren't you still passing around these objects? @sprillo what do you think about this?

richardyz98: I guess I'm just saying that `sample_lineage_event` would have to take additional parameters. But I think it might be helpful to show exactly what is being altered by `sample_lineage_event`. So I think I'm actually down to make them instance methods and not nested functions

mattjones315: Ah I see, I see. Yes, I think it's a bit cleaner but if it's a lot of work with few if any benefits I don't think it's needed.

richardyz98: I'll whip it up haha and then let's push!

mattjones315 (approved):
Looks fantastic! Let's merge it in!

PR #76, opened 2021-03-05
Ilp bug

mattjones315:
Hi team,

I've tracked down the bug in the `ILPSolver`. Despite our hunch that it had to do with missing data, I found that it was caused by duplicates in the character matrices. Essentially, I was passing duplicate targets to the `Gurobi` model which caused an error. I've fixed this bug in the `solve` method of the `ILPSolver`. I also have added two new tests that should hopefully cover this type of thing in the future -- 

```
TestILPSolver.test_ilp_solver_with_duplicates
TestILPSolver.test_potential_graph_inference_with_duplicates
```

I've also added a test to the ILPSolver test suite that checks that we correctly infer trees with missing data: `TestILPSolver.test_ilp_solver_missing_data`

Let me know what you two think!

### DIFF
--- a/cassiopeia/solver/ILPSolver.py
+++ b/cassiopeia/solver/ILPSolver.py
@@ -129,7 +129,7 @@ class ILPSolver(CassiopeiaSolver.CassiopeiaSolver):
             "|".join([str(r) for r in root]).encode("utf-8")
         ).hexdigest()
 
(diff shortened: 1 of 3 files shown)


richardyz98 (approved):
Looks great! Thanks for finding this bug Matt!

sprillo (approved):
LGTM

PR #78, opened 2021-03-12
[Ready for Review] Cas9-based Lineage Tracing Experiment Simulator

mattjones315:
This PR implements the Cas9 based lineage tracing data simulator. The purpose of this simulator is to provide a realistic framework for generating data akin to the technologies described in Chan et al, Nature 2019 and McKenna et al, Science 2016. This includes, but is not limited to:

* segmentation of characters into equally sized "cassettes"
* integration of specific mutation rates, to simulate the heavy-tailed Cas9-indel distributions observed in real studies
* inclusion of multiple missing data sources -- specifically, heritable missing data and stochastic missing data
* resection events that are caused by simultaneous Cas9 cutting on the same cassette, resulting in missing data

The character states are populated in a `CassiopeaiTree` that is passed in with branch lengths.

All tests are stored in `tests/simulator_tests/cas9_lineage_tracing_simulator_tests.py`.


cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,202 @@
...
+        """Silences cassettes.
+
+        Using the predefined silencing rate of this simulator, this function
+        simulates the rare event of transcriptional silencing.
+
+        Args:
+            character_array: Character array
+
+        Returns:
+            An updated character array.
```
sprillo, line 178: Actually this returns nothing: the `character_array` is modified in-place.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,202 @@
...
+
+        New states are added to the character array at the predefined cut
+        locations.
+
+        Args:
+            character_array: Character array
+            cuts: Loci being cut
+
+        Returns:
+            An updated character array.
```
sprillo, line 159: This method also seems to modify the `character_array` in-place; we should document this.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,202 @@
...
+        """
+
+        for i in cuts:
+            state = self.random.choice(list(self.mutation_priors.keys()), 1, p=list(self.mutation_priors.values()))
```
sprillo, line 163: This line is hard to understand, can we split it into multiple lines, each of which is easy to understand?

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,202 @@
...
+    """
+
+    def __init__(
+        self,
+        number_of_cassettes: int = 10,
+        size_of_cassette: int = 3,
+        mutation_rate: float = 0.01,
+        state_distribution: Callable[None, float] = lambda: np.random.exponential(1e-5),
+        number_of_states: int = 100,
+        silencing_rate: float = 1e4,
```
sprillo, line 65: 1e-4?

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,202 @@
...
+            p = 1 - (np.exp(-t * self.mutation_rate))
+    
+            character_array = tree.get_character_states(parent)
+            open_sites = [c for c in range(len(character_array)) if character_array[c] == 0]
+
+            number_of_mutations = np.random.binomial(len(open_sites), p)
+            new_cuts = np.random.choice(open_sites, number_of_mutations)
+            
+            # collapse cuts that are on the same cassette
+            character_array, cuts_remaining = self.collapse_sites(character_array, new_cuts)
```
sprillo, line 106: Hmm, this is not how I was expecting a double cut to happen. In my mind, whether a double cut happens or not should not depend on what cuts got _accumulated_ during the lifetime of a cell: a double cut should depend on whether the two cuts happened _close_ together in time.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,202 @@
...
+            parent = tree.get_parent(node) 
+            t = tree.get_time(node) - tree.get_time(parent)
+
+            p = 1 - (np.exp(-t * self.mutation_rate))
+    
+            character_array = tree.get_character_states(parent)
+            open_sites = [c for c in range(len(character_array)) if character_array[c] == 0]
+
+            number_of_mutations = np.random.binomial(len(open_sites), p)
+            new_cuts = np.random.choice(open_sites, number_of_mutations)
```
sprillo, line 103: Just to make sure: this is the same as tossing a coin with probability `p` for each open site?

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,202 @@
...
+    in which an entire cassette is transcriptionally silenced and therefore
+    not observed.
+
+    The function `overlay_data` will operate on the tree in place and will
+    specifically modify the data stored in the character attributes.
+
+    Args:
+        number_of_cassettes: Number of cassettes (i.e., arrays of target sites)
+        size_of_cassette: Number of editable target sites per cassette
+        mutation_rate: Exponential parameter for the Cas9 cutting rate.
```
sprillo, line 51: Have you considered having a different mutation rate for each site?

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,202 @@
...
+    resulting lineage will have |cassette|*(# of cassettes) characters.
+
+    Third, the class accepts a state distribution describing the relative
+    likelihoods of various indels. This is very useful, as it is typical that a
+    handful of mutations are far likelier than the bulk of the possible
+    mutations.
+
+    Finally, the class accepts a silencing rate. This is normally a rare event
+    in which an entire cassette is transcriptionally silenced and therefore
+    not observed.
```
sprillo, line 43: Maybe we should be more explicit that silencing happens when a cell divides? (As opposed to stochastic dropout)

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,202 @@
...
+        number_of_states: Number of states to simulate
+        silencing_rate: Silencing rate for the cassettes, per node
+    """
+
+    def __init__(
+        self,
+        number_of_cassettes: int = 10,
+        size_of_cassette: int = 3,
+        mutation_rate: float = 0.01,
+        state_distribution: Callable[None, float] = lambda: np.random.exponential(1e-5),
```
sprillo, line 63: Why a callable and not just a list of floats?

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,202 @@
...
+
+    The function `overlay_data` will operate on the tree in place and will
+    specifically modify the data stored in the character attributes.
+
+    Args:
+        number_of_cassettes: Number of cassettes (i.e., arrays of target sites)
+        size_of_cassette: Number of editable target sites per cassette
+        mutation_rate: Exponential parameter for the Cas9 cutting rate.
+        state_distribution: Distribution from which to simulate state
+            likelihoods
```
sprillo, line 53: So, this is not something that sums to 1? We pull out a bunch of values and then normalize?

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,202 @@
...
+        cassettes = self.get_cassettes()
+        cut_site_by_cassette = np.digitize(range(len(character_array)), cassettes)
+        for cassette in cassettes:
+            if np.random.uniform() < self.silencing_rate:
+                indices = np.where(cut_site_by_cassette)
+                left, right = np.min(indices), np.max(indices)
+                for site in range(left, right+1):
+                    character_array[site] = -1
+
+    def get_cassettes(self) -> List[int]:
```
sprillo, line 190: Should this (and other methods) be made private?

sprillo (commented):
I know this is a WIP, but since it was short I took a look out of curiosity and added some comments!

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
richardyz98: yes!

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: Good catch! I changed this, and addressed the comment below by creating a copy that we modify and returning this modified copy.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: See response above.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: I fixed this in a recent commit, how does it look now?

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: Good catch!

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: I totally agree - this is not necessarily _exactly_ how double cuts would happen.

However, there are lots of other things that we are not modeling. We're not actually modeling DNA repair, we're not actually modeling how long Cas9 takes to induce a cut, etc. These all undoubtedly play a part in these multiple-site resections. I think a good approximation is to assume that if Cas9 cuts at two sites on the same recorder in a cell's lifetime, we'll introduce a resection event.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: We have done this experiment before -- you can see it in the "Phased Editor" part of our Genome Biology paper. We could easily add this to the simulator.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: Good call -- I'll add this to the documentation.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: Ah, that'd be nice. I'll add an option for this!

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: Yes, this is exactly how I am using it

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: I prefer to keep these public for testing purposes. Is there a strong reason for why we'd not like a user to have access to these methods?

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,327 @@
...
+
+            # introduce new states at cut sites
+            character_array = self.introduce_states(
+                character_array, cuts_remaining
+            )
+
+            # silence cassettes
+            character_array = self.silence_cassettes(
+                character_array, self.heritable_silencing_rate
+            )
```
richardyz98, line 202: Is it true that silencing only happens at division events?

richardyz98 (changes requested):

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,327 @@
...
+        for node in tree.nodes:
+            character_matrix[node] = [-1] * number_of_characters
+
+        for node in tree.depth_first_traverse_nodes(tree.root, postorder=False):
+
+            if tree.is_root(node):
+                character_matrix[node] = [0] * number_of_characters
+                continue
+
+            parent = tree.parent(node)
```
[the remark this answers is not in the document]
mattjones315: Sorry, I don't see how this can happen -- I'm checking specifically above whether or not the node is the root. You should never hit `parent = tree.parent(node)` if the node is the root.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,327 @@
...
+
+                if np.random.uniform() < p:
+                    new_cuts.append(site)
+
+            # collapse cuts that are on the same cassette
+            cuts_remaining = new_cuts
+            if self.size_of_cassette > 1:
+                character_array, cuts_remaining = self.collapse_sites(
+                    character_array, new_cuts
+                )
```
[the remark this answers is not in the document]
mattjones315: I was thinking about this too, actually. But if a user does not want this behavior, they should just pass `size_of_cassette = 1`

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: Yes and no. It might as well just happen at division events in the simulation -- because if the silencing happens at all during a cell's lifetime it's going to get passed on to its children. But more generally, we don't really know when the heritable silencing happens.

mattjones315:
Hey @richardyz98, I believe I hit most of your comments (either by responding to them or by changing stuff in the code). Thanks for the feedback!

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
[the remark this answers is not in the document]
richardyz98: Ah that is true! Perhaps we can mention that in the docstring then

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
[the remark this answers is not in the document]
richardyz98: I missed this! Thank you for clarifying

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
richardyz98: I guess what changes is our interpretation of the `heritable_silencing_rate` parameter. The way it's implemented now, it's dependent on the underlying tree topology (how many generations occur determines the probability that a silencing event will happen in any given character in any given cell). Whereas if we used the silencing rate as a true rate in an exponential process like we're doing for the mutation rate, it might be easier to build estimators for. (If a silencing event happens at all on a branch, then it'll just override any other mutation that might have occurred at a site on that branch in the case where we use an exponential).

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,328 @@
...
+                raise DataSimulatorError(
+                    "Mutation rate needs to be" " non-negative."
+                )
+            self.mutation_rate_per_character = mutation_rate
+
+        self.mutation_priors = state_priors
+        if self.mutation_priors is None:
+            self.mutation_priors = {}
+            probabilites = [
+                state_generating_distribution() for _ in range(number_of_states)
```
sprillo, line 132: This step happens before a random seed is set, so it might produce different results on different runs of the global script? I would probably move this code out of the `__init__` method and into the `overlay_data` method.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,328 @@
...
+            specified to the simulator.
+        number_of_states: Number of states to simulate
+        mutation_priors: A mapping from state to probability that a user
+            can specify. If this argument is not None, states will not be
+            pulled from the state distribution.
+        heritable_silencing_rate: Silencing rate for the cassettes, per node,
+            simulating heritable missing data events.
+        stochastic_silencing_rate: Rate at which to randomly drop out cassettes,
+            to simulate dropout due to low sensitivity of assays.
+        random_seed: Random seed to use for deterministic simulations. Note
```
sprillo, line 70: I think we should be more precise and say that this sets the *numpy* random seed.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,328 @@
...
+            open_sites = [
+                c
+                for c in range(len(character_array))
+                if character_array[c] == 0
+            ]
+
+            new_cuts = []
+            for site in open_sites:
+                mutation_rate = self.mutation_rate_per_character[site]
+                p = 1 - (np.exp(-t * mutation_rate))
```
sprillo, line 183: Can we call this variable `p` something more expressive, like `mutation_probability`?

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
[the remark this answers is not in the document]
sprillo: What if a user want to have different cut rates for each site of the cassette (a feature I requested in a comment above), and doesn't want collapsing of cuts? In this case, I like @richardyz98 's proposal to add a parameter that allows the user to specify avoiding this behavior.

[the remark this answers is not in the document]
sprillo: Also, I mentioned this in my previous review, but I don't think this model of collapsing is biologically accurate. I am fine leaving it as it is, as long as we explain exactly how collapsing is handled: regardless of when then cuts happen during the lifetime of the cell (i.e. even if two cuts happen a very long time apart), collapsing will still happen when the cell divides.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
sprillo: Nice catch @richardyz98 . I agree with @richardyz98 , the `heritable_silencing_rate` is per division event and not per unit time. @mattjones315 which is more reasonable biologically? We should implement the more biologically reasonable one. I feel like per time unit (as @richardyz98 mentioned) is more reasonable: the longer the cell lives, the more likely a heritable dropout is.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,328 @@
...
+                character_array, self.heritable_silencing_rate
+            )
+
+            character_matrix[node] = character_array
+
+        # apply stochastic silencing
+        for leaf in tree.leaves:
+            character_matrix[leaf] = self.silence_cassettes(
+                character_matrix[leaf], self.stochastic_silencing_rate
+            )
```
sprillo, line 211: I think it would be *great* if we allowed the user to specify what integers to use to represent stochastic and heritable dropouts. That way, we can easily simulate data where we know whether a dropout in the character matrix is heritable or stochastic. This is really easy to implement: we just have to add two more arguments to `__init__`. I think as long as we require that they are negative numbers, and by default they are `-1`, we should be good. I know that some downstream methods assume that the missing state is `-1`, so we should definitely add a warning.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
sprillo: No strong reason, other than when reading code I like being able to distinguish key methods (public) from implementation details (private). But I can also infer this by looking at the abstract base class we are implementing, so not really an issue. In this case, you are right that there might be a reason for a user to want to use this method.

sprillo (changes requested):
Added some comments (mostly feature requests). Let me know what you think!

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
sprillo: Sounds good! So long as we document this precisely.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,328 @@
...
+
+from cassiopeia.data import CassiopeiaTree
+from cassiopeia.simulator.DataSimulator import DataSimulatorError
+from cassiopeia.simulator import LineageTracingDataSimulator
+
+
+class Cas9LineageTracingDataSimulator(LineageTracingDataSimulator):
+    """Simulates Cas9-based lineage tracing data.
+
+    This subclass of `LineageTracingDataSimulator` extends the `overlay_data`
```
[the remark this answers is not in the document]
mattjones315: Good call -- I'll change this in the documentation!

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,328 @@
...
+    The class accepts several parameters that govern the Cas9-based tracer.
+    First and foremost is the cutting rate, describing how fast Cas9 is able to
+    cut. We model Cas9 cutting as an exponential process, parameterized by the
+    specified mutation rate - specifically, for a lifetime `t` and parameter
+    `lambda`, the expected probability of Cas9 mutation, per site, is
+    `exp(-lambda * t)`.
+    
+    Second, the class accepts architecture of the recorder - described by the
+    size of the cassette (by default 3) and the number of cassettes. The
+    resulting lineage will have |cassette|*(# of cassettes) characters.
```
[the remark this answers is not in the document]
mattjones315: Good call

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,328 @@
...
+        heritable_silencing_rate: float = 1e-4,
+        stochastic_silencing_rate: float = 1e-2,
+        random_seed: Optional[int] = None,
+    ):
+
+        if number_of_cassettes <= 0 or not isinstance(number_of_cassettes, int):
+            raise DataSimulatorError("Specify a positive number of cassettes.")
+        if size_of_cassette <= 0 or not isinstance(size_of_cassette, int):
+            raise DataSimulatorError(
+                "Specify a positive number of cut-sites" " per cassette."
```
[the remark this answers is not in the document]
mattjones315: To keep line lengths at 80 characters.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,328 @@
...
+            )
+
+        self.size_of_cassette = size_of_cassette
+        self.number_of_cassettes = number_of_cassettes
+
+        if isinstance(mutation_rate, float):
+            if mutation_rate < 0:
+                raise DataSimulatorError(
+                    "Mutation rate needs to be" " non-negative."
```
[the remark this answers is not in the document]
mattjones315: Same as above, re: line length.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,328 @@
...
+
+    Raises:
+        DataSimulatorError if assumptions about the system are broken.
+    """
+
+    def __init__(
+        self,
+        number_of_cassettes: int = 10,
+        size_of_cassette: int = 3,
+        mutation_rate: Union[float, List[float]] = 0.01,
```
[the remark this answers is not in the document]
mattjones315: You probably just missed this, but I do support this by allowing the user to pass in a List of floats. Each entry corresponds to a mutation rate for a given character. This gives the user ultimate control over the mutation rates, and can be organized according to cassette, target-site per cassette, or something else more clever.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,328 @@
...
+                self.number_of_cassettes * self.size_of_cassette
+            ):
+                raise DataSimulatorError(
+                    "Length of mutation rate array is not"
+                    " the same as the number of characters."
+                )
+
+            if np.any(np.array(mutation_rate) < 0):
+                raise DataSimulatorError(
+                    "Mutation rate needs to be" " non-negative."
```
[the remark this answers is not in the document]
mattjones315: Same as above, re: line length.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: Ah, this is a great point! I'll move this to the `overlay_data` method.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: Good point!

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: This is a good point Richard, I've changed the variable name in my latest set of commits.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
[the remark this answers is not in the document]
mattjones315: There are two points here --

(1) **Cutting rates & an options for collapsing**. This code supports both of these with a lot of flexibility. A user can pass in an arbitrary list of mutation rates for their lineage tracer by passing in a list of floats to `mutation_rate`. This allows a user to specify the mutation rates with exquisite control -- per target site, per cassette, or a combination of both. Re: collapsing -- a user can encode this, *without changing any other behavior*, by specifying the `size_of_cassette` to be 1. You'll still get transcriptional silencing here but no resections.

(2) Your point is well-taken Sebastian. My point is that the biology behind double-resections is not well known at all, especially with respect to how long the Cas9 is residing while cutting and what the time window is. We cannot feasibly model everything that happens biologically; e.g.the random mutations that are introduced upon cell division, the opening and closing of DNA regions, the silencing of *guides*, etc. I think that this is a fair generalization to how resections are taking place. This is a general comment on the relationship between simulation & reality that I will also make below when we talk about the transcriptional silencing of the cassettes. In any event, I will make an addendum to the documentation to make this more clear!

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: This is related to a point I made above about the intersection between biology and the simulation -- I'm actually not entirely sure when the transcriptional silencing would happen (i.e., if it's at a division event or if it could occur during a cell's lifetime). I agree though, for the purposes of building estimators, let's encode this as a time-dependent process. I'm happy to add this to the simulator!

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: That's a good idea! I'll add this to the simulator.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: Okay, sounds good! Let's keep them public for now.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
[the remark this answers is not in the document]
sprillo: Oh, I see, I didn't notice the user could specify a list of floats for the `mutations_rate`! We should update the docstring to clarify this, currently it says: `mutation_rate: Exponential parameter for the Cas9 cutting rate.`

I'm cool with the double-resections logic then!

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
sprillo: Gotcha, then I'm fine either way! Let's just remember to document it.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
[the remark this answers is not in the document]
sprillo: Isn't:
`"Specify a positive number of cut-sites" " per cassette."` equivalent to:
`"Specify a positive number of cut-sites per cassette."` ?
I'm still confused about this.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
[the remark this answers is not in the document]
sprillo: Yeah, I totally missed this Matt! As I mentioned above, let's make this clear in the docstring for the `mutation_rate` parameter.

richardyz98 (commented):
Hi Matt,
I think it's looking great now!

Did you see my comment in the slack by chance? Perhaps in a future iteration we can have it so that we can allow the simulator to track the ground truth mutations so that we can see what dropped out mutations should have corresponded to? This would be really helpful in the missing data work

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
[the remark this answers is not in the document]
mattjones315: great!

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: In my latest commit I've updated the tree simulator to have a time-dependence to the transcriptional silencing. here, I think it makes sense to model -- especially because we'll want to use some estimators here

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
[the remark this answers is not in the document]
mattjones315: You just can't see it here in the Github rendering, but in the real code there's a line separating these two strings. So, the code is formatted to respect the 80 character limit.

mattjones315:
Hi @sprillo and @richardyz98 --

I believe I've addressed all of your comments. These were all really helpful, so thank you!

Richard: regarding the missing data, I think this might introduce too much engineering into the simulation and also has consequences with respect to the lineage tracing simulator. I could see how we would do it, but I propose that we have a separate simulator that will "corrupt" the data for your purposes. In particular, you would simulate the data onto the tree with no collapsing and no missing data and then have a separate simulation process that would update the character matrix by introducing missing data. Would this work for your purpose?

If you think it would be a lot cleaner to have this baked into the current simulator, I think I know how to introduce this cleanly.

richardyz98:
Hi Matt,
I think I know what you mean by adding a new simulator that "corrupts" existing trees with missing data. I think that might be a useful functionality to have, so I'll approve this and implement that later for my own purposes. Thanks for all of the great work here!

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,367 @@
...
+    cassettes due to the low sensitivity of the RNA-sequencing assay. 
+
+    The function `overlay_data` will operate on the tree in place and will
+    specifically modify the data stored in the character attributes.
+
+    Args:
+        number_of_cassettes: Number of cassettes (i.e., arrays of target sites)
+        size_of_cassette: Number of editable target sites per cassette
+        mutation_rate: Exponential parameter for the Cas9 cutting rate. The user
+            can optionally pass in a float, in which every site will mutate at
```
sprillo, line 70: I don't think this is `optionally`, I would say `either` instead?

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,367 @@
...
+            updated_character_array[i] = state
+
+        return updated_character_array
+
+    def silence_cassettes(
+        self, character_array: List[int], silencing_rate: float, missing_state: int = -1
+    ) -> List[int]:
+        """Silences cassettes.
+
+        Using the specied silencing rate, this function will randomly select
```
sprillo, line 323: typo: `specied`

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -0,0 +1,367 @@
...
+                raise DataSimulatorError(
+                    "Length of mutation rate array is not"
+                    " the same as the number of characters."
+                )
+
+            if np.any(np.array(mutation_rate) < 0):
+                raise DataSimulatorError(
+                    "Mutation rate needs to be" " non-negative."
+                )
+            self.mutation_rate_per_character = mutation_rate
```
sprillo, line 148: Let's make a copy to avoid exposing the internal representation of the class (similar to the issue we had with CassiopeiaTree). E.g.: `self.mutation_rate_per_character = mutation_rate[:]`

mattjones315: Good call, this is done in my latest commit.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
mattjones315: This is what I meant in this context; happy to change it to "either" for clarity

sprillo: Oh, I see! I was thinking in the context of python types: `mutation_rate` is a `Union` rather than an `Optional`. I'm glad to leave as it is though, I'm just overloaded with mypy!

mattjones315: Haha, no worries! I changed it to "either" for clarity.

PR #81, opened 2021-03-18
added branch length to newick generation and itol plotting

richardyz98:
Hi,
Here I've implemented the functionality for to_newick to add branch lengths to the outputted string, and also fixed some logic in the main itol plotting function to make it handle taking in an api_key and project_name correctly.

Thanks!

### DIFF
--- a/cassiopeia/data/CassiopeiaTree.py
+++ b/cassiopeia/data/CassiopeiaTree.py
@@ -776,7 +776,7 @@ class CassiopeiaTree:
         """Returns newick format of tree.
         
         Args:
-            record_edge_weights: Whether to record branch lengths on the tree
+            record_branch_lengths: Whether to record branch lengths on the tree
             in the newick string
 
         Returns:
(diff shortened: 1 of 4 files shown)


cassiopeia/data/CassiopeiaTree.py
```diff
@@ -772,9 +772,17 @@ def leaves_in_subtree(self, node) -> List[str]:
...
         ]
 
-    def get_newick(self) -> str:
-        """Returns newick format of tree."""
-        return utilities.to_newick(self.__network)
+    def get_newick(self, record_branch_lengths = False) -> str:
+        """Returns newick format of tree.
+        
+        Args:
+            record_edge_weights: Whether to record branch lengths on the tree
```
mattjones315, line 779: documentation is off -- this should be `record_branch_lengths`.

mattjones315 (commented):
Looks good! Just need to change that one little documentation issue.

mattjones315 (approved):
thanks for fixing this, let's merge it in!

PR #83, opened 2021-03-23
[Ready for Review] Add UniformLeafSubsampler and moved collapsing mutationless edges into CassiopeiaTree

richardyz98:
Hi all,
This is just a small PR moving collapsing mutationless edges into the CassiopeiaTree.

Also this brings up the question: in `get_mutations_on_edge`, should we report mutations from a non-0 character to the missing state. The current implementation does not show these. If so, I'll use `get_mutations_on_edge` in the implementation of `collapse_mutationless_edges`, otherwise I'll continue to match the character vectors.

Thanks in advance!
Richard


mattjones315 (changes requested):
Looking good, Richard!

All the tests pass which is nice and the code is well-written.

I have some comments that I'd like to get addressed before merging in:

(1) There are some places where I'd like to see some more documentation. There should be file headers on every file, for the most part, so let's add general descriptions to the new files. There are also some places where the documentation can be clearer, especially around what is and what is not preserved in the subsampling and what the preferred parameters are.

(2) In the collapse mutationless edges, we rely too much on the `networkx` functionality. Whenever possible I think we should be using `CassiopeiaTree` APIs - that means using `self.children` instead of `self.__network.successors(node)` for example. I also think we should add private methods that are capable of adding/removing edges/nodes from the tree and using these new `CassiopeiaTree` functions internally to do these topology modifications. This will make it easier to update if we ever move to a different backend.

Thanks again Richard!

cassiopeia/simulator/UniformLeafSubsampler.py
```diff
@@ -0,0 +1,72 @@
...
+
+class UniformLeafSubsampler(LeafSubsampler):
+    def __init__(
+        self, ratio: Optional[float] = None, n_leaves: Optional[int] = None
+    ):
+        """
+        Uniformly subsample leaf samples.
+
+        If 'ratio' is provided, samples 'ratio' of the leaves, rounded down,
+        uniformly at random. If instead 'n_leaves' is provided, 'n_leaves' of
```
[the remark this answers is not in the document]
richardyz98: I think I provide an explicit test in the constructor to ensure that only 1 is being passed

cassiopeia/simulator/UniformLeafSubsampler.py
```diff
@@ -0,0 +1,72 @@
...
+        if ratio is None and n_leaves is None:
+            raise LeafSubsamplerError(
+                "At least one of 'ratio' and 'n_leaves' " "must be specified."
+            )
+        if ratio is not None and n_leaves is not None:
+            raise LeafSubsamplerError(
+                "Exactly one of 'ratio' and 'n_leaves'" "must be specified."
+            )
+        self.__ratio = ratio
+        self.__n_leaves = n_leaves
```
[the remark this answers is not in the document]
richardyz98: I based this off of what Sebastian was doing. I think the belief is that once we establish the `ratio` and `n_leaves` inputs, we don't want the user to be able to go in and alter those. I think this is just for consistency and to avoid confusion, since if the user wanted to specify different `ratio` and `n_leaves`  they would easily create a new subsampler object. Additionally, we're checking to see if both `ratio` and `n_leaves` are specified in the constructor (if they are, we error), and we don't want the user to initialize the object and then add either of these afterwards

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -645,9 +737,34 @@ def set_branch_length(self, parent: str, child: str, length: float):
...
-        self.__network[parent][child]["length"] = length
+        self.__set_branch_length(parent, child, length)
 
-        for (u, v) in self.depth_first_traverse_edges(source=parent):
+        for u, v in self.depth_first_traverse_edges(source=parent):
+            self.__network.nodes[v]["time"] = (
+                self.__network.nodes[u]["time"] + self.__network[u][v]["length"]
+            )
+
+    def set_branch_lengths(self, branch_length_dict: Dict[Tuple[str, str], float]) -> None:
```
mattjones315, line 747: Some places you have return `None`, some places you do not. Either is fine (I normally omit them), but can we make this consistent?

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -645,9 +737,34 @@ def set_branch_length(self, parent: str, child: str, length: float):
...
+        Args:
+            branch_dict: A dictionary of edges to updated branch lengths
+        """
+        self.__check_network_initialized()
+
+        for edge, length in branch_length_dict.items():
+            u, v = edge[0], edge[1]
+            if length < 0:
+                raise CassiopeiaTreeError("Edge length must be positive.")
+            self.__network[u][v]["length"] = length
```
mattjones315, line 765: Why don't we use `set_branch_length` here?

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -371,7 +366,7 @@ def leaves(self) -> List[str]:
 
         if "leaves" not in self.__cache:
             self.__cache["leaves"] = [
-                n for n in self.__network if self.__network.out_degree(n) == 0
+                n for n in self.__network if self.out_degree(n) == 0
```
mattjones315, line 369: Why not just use `is_leaf` here instead of `self.out_degree`?

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -389,7 +384,7 @@ def internal_nodes(self) -> List[str]:
 
         if "internal_nodes" not in self.__cache:
             self.__cache["internal_nodes"] = [
-                n for n in self.__network if self.__network.out_degree(n) > 1
+                n for n in self.__network if self.out_degree(n) > 1
```
mattjones315, line 387: Why not use `is_internal_node` here instead of `self.out_degree(n) > 1`?

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -517,6 +513,88 @@ def children(self, node: str) -> List[str]:
...
+            node: A node in the tree.
+            
+        Raises:
+            CassiopeiaTreeError if the tree is not initialized.
+        """
+
+        self.__check_network_initialized()
+
+        return self.__network.in_degree(node)
+
```
mattjones315, line 597: Why do we need these utilities? It looks like you're using them to check for roots and leaves, but we already have functions to do this -- `is_root`, `is_leaf`, and `is_internal_node`. If there's not another purpose for them, I propose we remove them.

(Also - if you're curious about the out degree, you can get this by len(self.children(n))`. By definition, all nodes should have _at most_ 1 in degree (the exception being the root))

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -772,7 +889,7 @@ def leaves_in_subtree(self, node) -> List[str]:
         return [
             n
             for n in self.depth_first_traverse_nodes(source=node)
-            if self.__network.out_degree(n) == 0
+            if self.out_degree(n) == 0
```
mattjones315, line 892: Why not use `is_leaf`?

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -876,71 +993,92 @@ def relabel_nodes(self, relabel_map: Dict[str, str]) -> None:
...
+        if not self.is_leaf(node):
+            raise CassiopeiaTreeError("Node is not a leaf.")
+
+        if self.in_degree(node) > 0:
+            curr_parent = self.parent(node)
+            self.__remove_node(node)
             while (
-                self.__network.out_degree(curr_parent) < 1
-                and self.__network.in_degree(curr_parent) > 0
+                self.out_degree(curr_parent) < 1
```
mattjones315, line 1021: `self.out_degree(curr_parent)` -> `len(self.children(curr_parent)`

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -876,71 +993,92 @@ def relabel_nodes(self, relabel_map: Dict[str, str]) -> None:
...
+                and self.in_degree(curr_parent) > 0
```
mattjones315, line 1022: `self.in_degree(curr_parent) > 0` -> `not self.is_root(curr_parent)`

cassiopeia/simulator/UniformLeafSubsampler.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class UniformLeafSubsampler(LeafSubsampler):
+    def __init__(
+        self, ratio: Optional[float] = None, n_leaves: Optional[int] = None
+    ):
+        """
+        Uniformly subsample leaf samples.
+
+        If 'ratio' is provided, samples 'ratio' of the leaves, rounded down,
```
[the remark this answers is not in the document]
mattjones315: Did this get addressed? I think you address it at around line 75, but please confirm.

mattjones315 (changes requested):
Hi Richard,

Looking good! I left a few comments, mainly in the CassiopeiaTree class. I think we can write some cleaner code that leverages existing API calls. Let me know if you have any other questions.

Thanks!

cassiopeia/data/CassiopeiaTree.py
richardyz98: `set_branch_length` does a consistency operation, so we'd have to do a consistency operation for each branch. It's more efficient for bulk branch setting I think to do just one consistency operation across the entire tree after all the branches are added. I should be using `__set_branch_length` though you are correct

cassiopeia/simulator/UniformLeafSubsampler.py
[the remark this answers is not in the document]
richardyz98: Yes this was addressed!

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -348,7 +348,7 @@ def root(self) -> str:
 
         if "root" not in self.__cache:
             self.__cache["root"] = [
-                n for n in self.__network if self.in_degree(n) == 0
+                n for n in self.__network if self.__network.in_degree(n) == 0
```
mattjones315: For consistency, why don't we just use `[n for n in self.__network if self.is_root(n)]`?

mattjones315 (commented):
Looking good! I see just one place we should make a change in the `CassiopeiaTree` class - I'll go ahead and do this myself. Otherwise, I think this should be good to merge in!

mattjones315 (approved):

PR #84, opened 2021-03-26
[Don't review] Add UniformCellSubsampler

sprillo:
Pushing this to a branch at Richard's request.

### DIFF
--- a/cassiopeia/simulator/CellSubsampler.py
+++ b/cassiopeia/simulator/CellSubsampler.py
@@ -1,0 +1,124 @@
+import abc
+import networkx as nx
+import numpy as np
+from typing import Optional
+
+from cassiopeia.data import CassiopeiaTree
+
+
+class CellSubsamplerError(Exception):
+    """An Exception class for the CellSubsampler class."""
+
+    pass
+
+
+class CellSubsampler(abc.ABC):
+    r"""
+    Abstract base class for all cell samplers.
+
+    A CellSubsampler implements a method 'subsample' which, given a Tree,
+    returns a second Tree which is the result of subsampling cells
+    (i.e. leafs) of the tree. Only the tree topology will be created for the
+    new tree.
+    """
+
+    @abc.abstractmethod
+    def subsample(self, tree: CassiopeiaTree) -> CassiopeiaTree:
+        r"""
+        Returns a new CassiopeiaTree which is the result of subsampling
+        the cells in the original CassiopeiaTree.
+
+        Args:
+            tree: The tree for which to subsample leaves.
+        """
+
+
+class UniformCellSubsampler(CellSubsampler):
+    def __init__(
+        self, ratio: Optional[float] = None, n_cells: Optional[int] = None
+    ):
+        r"""
+        Uniformly subsample cells.
+
+        If 'ratio' is provided, samples 'ratio' of the leaves, rounded down,
+        uniformly at random. If instead 'n_cells' is provided, 'n_cells' of
+        the leaves are sampled uniformly at random. Only one of the two
+        criteria can be provided.
+        """
+        if ratio is None and n_cells is None:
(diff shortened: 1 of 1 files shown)


cassiopeia/simulator/CellSubsampler.py
```diff
@@ -0,0 +1,124 @@
...
+    """An Exception class for the CellSubsampler class."""
+
+    pass
+
+
+class CellSubsampler(abc.ABC):
+    r"""
+    Abstract base class for all cell samplers.
+
+    A CellSubsampler implements a method 'subsample' which, given a Tree,
```
mattjones315, line 19: `Tree` -> `CassiopeiaTree`.

What information does this subsampler preserve? If this is done _before_ a tree has been populated, do we just subsample the character matrix?

If the tree has been populated, do we modify both the character matrix and the tree?

cassiopeia/simulator/CellSubsampler.py
```diff
@@ -0,0 +1,124 @@
...
+    returns a second Tree which is the result of subsampling cells
```
mattjones315, line 20: ditto above -- `Tree` -> `CassiopeiaTree`.

cassiopeia/simulator/CellSubsampler.py
```diff
@@ -0,0 +1,124 @@
...
+        r"""
+        Returns a new CassiopeiaTree which is the result of subsampling
+        the cells in the original CassiopeiaTree.
+
+        Args:
+            tree: The tree for which to subsample leaves.
+        """
+
+
+class UniformCellSubsampler(CellSubsampler):
```
mattjones315, line 36: Should we make a new file for this class? I presume we'll have different CellSubsamplers?

Because if not, we should just make this part of the `CassiopeiaTree` class I think. (In the form of having a method dubbed `subsample` that takes in some fraction and randomly downsamples the tree)

cassiopeia/simulator/CellSubsampler.py
```diff
@@ -0,0 +1,124 @@
...
+                "At least one of 'ratio' and 'n_cells' " "must be specified."
+            )
+        if ratio is not None and n_cells is not None:
+            raise CellSubsamplerError(
+                "Exactly one of 'ratio' and 'n_cells'" "must be specified."
+            )
+        self.__ratio = ratio
+        self.__n_cells = n_cells
+
+    def subsample(self, tree: CassiopeiaTree) -> CassiopeiaTree:
```
mattjones315, line 59: Can you add a docstring here?  We'll need it for documentation on the website.

cassiopeia/simulator/CellSubsampler.py
```diff
@@ -0,0 +1,124 @@
...
+                induced_subtree_degs[node] >= 2
+                or induced_subtree_degs[node] == 0
+            ):
+                nodes.append(node)
+                edges.append((up[node], node))
+        subtree_topology = nx.DiGraph()
+        subtree_topology.add_nodes_from(nodes)
+        subtree_topology.add_edges_from(edges)
+        res = CassiopeiaTree(
+            tree=subtree_topology,
```
mattjones315, line 120: What about the character matrix that was in the original CassiopeiaTree?

And what about other tree-specific attributes like fitness etc?

cassiopeia/simulator/CellSubsampler.py
```diff
@@ -0,0 +1,124 @@
+import abc
```
mattjones315, line 1: Let's add a file header here.

mattjones315 (changes requested):
Hey Sebastian, this is looking nice!

I added my review here because it looked like you requested it (despite it saying "Don't Review" in the PR name). Either way, I'm happy to take a look again after some of these comments have been resolved and you are ready!

PR #85, opened 2021-03-29
fix sort_cellranger_bam chunked sorting

Lioscro:
Hi!
I've been using Cassiopeia to preprocess BAM files that were manually generated (manually tagged with CB and UR; all reads unmapped). The BAM file contains 10729723 reads.
When `cassiopeia.pp.UMI_utils.sort_cellranger_bam` is called from `cassiopeia.pp.collapse_umis`, the chunk loop is not appropriately indented, so only the last chunk is processed.
https://github.com/YosefLab/Cassiopeia/blob/30907209afefbbf353383d8fea30708d7b43ac7d/cassiopeia/preprocess/UMI_utils.py#L105-L118

In my case, only the last 729723 reads are processed (and the other chunk with 10M reads is discarded). This PR fixes this issue.

Additionally, I've added `check_sq=False` to https://github.com/YosefLab/Cassiopeia/blob/30907209afefbbf353383d8fea30708d7b43ac7d/cassiopeia/preprocess/UMI_utils.py#L98
to allow unmapped BAMs.

### DIFF
--- a/cassiopeia/preprocess/UMI_utils.py
+++ b/cassiopeia/preprocess/UMI_utils.py
@@ -1,7 +1,7 @@
 """
 This file contains all functions pertaining to UMI collapsing and preprocessing.
-Invoked through pipeline.py and supports the collapseUMIs and 
-errorCorrectUMIs functions. 
+Invoked through pipeline.py and supports the collapseUMIs and
+errorCorrectUMIs functions.
 """
 import os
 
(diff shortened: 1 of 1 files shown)


mattjones315:
This is fantastic, and thanks so much for catching this @Lioscro! This must have gotten un-indented during our porting over of old code to a new file.

Regarding the `check_sq=False` change -- I presume this makes it a lot easier for users who will be creating their own BAM files like you're doing? Is there any reason why we'd want to enforce `check_sq = True`, or to add this as an argument to the function?

Lioscro:
Yes, it simply disables the check for SQ entries in the BAM header, which do not exist in my case because the BAM is entirely of unaligned sequences. It removes the need to add dummy sequences as a workaround, while it doesn't change behavior for BAMs with sequences.

mattjones315:
Makes sense -- this is a good fix for your situation and shouldn't affect the more traditional use of this pipeline. Thanks! 

Look ready to merge in.

PR #89, opened 2021-04-29
Update iTOL plotting function + minor fix

Lioscro:
I've added a new argument `files` to `upload_and_export_itol` function that can be used to upload additional custom datasets. Also, any keyword arguments provided to the function are now considered as iTOL export parameters. This change makes it easier to customize the iTOL export a bit more than is currently allowed.

Also, a small fix where a dataframe was not being sorted inplace.

### DIFF
--- a/cassiopeia/plotting/itol_utilities.py
+++ b/cassiopeia/plotting/itol_utilities.py
@@ -96,6 +96,8 @@ def upload_and_export_itol(
         random_state: A random state for reproducibility
         files: Additional files to upload
         verbose: Include extra print statements.
+        **kwargs: additional keyword arguments are passed as iTOL export parameters.
+            See https://itol.embl.de/help.cgi#batch for a full list.
 
     Raises:
         iTOLError if iTOL credentials cannot be found, if the output format is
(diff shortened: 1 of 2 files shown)


cassiopeia/plotting/itol_utilities.py
```diff
@@ -45,7 +45,9 @@ def upload_and_export_itol(
     use_branch_lengths: bool = False,
     palette: List[str] = palettes.Category20[20],
     random_state: Optional[np.random.RandomState] = None,
+    files: Optional[List[str]] = None,
```
mattjones315, line 48: Let's name this something more informative, like `user_plotting_files`. Happy to entertain better names, but `files` is too obscure, I think.

cassiopeia/preprocess/pipeline.py
```diff
@@ -570,7 +570,7 @@ def filter_molecule_table(
     t0 = time.time()
     logging.info("Begin filtering reads...")
     input_df["status"] = "good"
-    input_df.sort_values("readCount", ascending=False)
+    input_df.sort_values("readCount", ascending=False, inplace=True)
```
mattjones315, line 573: Ah, good catch! Thanks.

mattjones315 (commented):
Nice PR! Just a small comment that I think we should use a more clear argument name than `files`. Thanks!

mattjones315 (approved):
Looks great!

PR #90, opened 2021-05-11
remove pathlib from requirements and set python version

Lioscro:
Pypi's `pathlib` is no longer being maintained and causes an error in Python 3.8. Instead, `pathlib` is included in the standard library starting from Python 3.4. Since we require users to have at least Python 3.6 installed (in the README), installing the pypi version of this library is undesirable.
See https://pypi.org/project/pathlib/

In line with this, I've added Python>=3.6 as a requirement in `setup.py`.
Also, fixed a typo from `test_require` to `tests_require`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,6 @@ requirements = [
     "ete3 >= 3.1.1",
     "argparse >= 1.1",
     "Biopython >= 1.71",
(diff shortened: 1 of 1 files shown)


mattjones315 (approved):
LGTM!

PR #91, opened 2021-05-11
Leaf subsampler border case bug

sprillo:
Hey @mattjones315 @richardyz98 ! I was using the `UniformLeafSubsampler` in my simulations and I ran into a bug. When no cells are subsampled, the tree should not change, and yet it was changing for me. I added a failing test called `test_no_subsampling` that reproduces the issue. I think the problem is simply that `collapse_unifurcations` is removing the unique edge that hangs from the root in my trees; recall that the root has degree 1 in my cell lineages.


richardyz98:
Closing because was resolved, additional documentation added in #135 to avoid future confusion

PR #94, opened 2021-05-13
added STDR wrap-around

richardyz98:
Hi Matt, here's what I have for the STDR wrap-around currently. It's not thoroughly documented, so I'll work on that.

### DIFF
--- a/cassiopeia/solver/STDRSolver.py
+++ b/cassiopeia/solver/STDRSolver.py
@@ -1,0 +1,35 @@
+from typing import Callable
+import dendropy
+import numpy as np
+import ete3
+import spectraltree
+from typing import Callable
+
+from cassiopeia.data import CassiopeiaTree
+from cassiopeia.data import utilities as data_utilities
+from cassiopeia.solver import CassiopeiaSolver
+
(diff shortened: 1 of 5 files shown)


cassiopeia/solver/STDRSolver.py
```diff
@@ -0,0 +1,35 @@
...
+
+        stdr_nj = spectraltree.STDR(spectraltree.NeighborJoining, self.similarity_function)   
+
+        tree_stdr_nj = stdr_nj(character_matrix.values, 
+                taxa_metadata= metadata, 
+                threshold = 64,
+                min_split = 1,
+                merge_method = "least_square", 
+                verbose=False)
+        tree = data_utilities.ete3_to_networkx(ete3.Tree(tree_stdr_nj.as_string(schema="newick")))
```
richardyz98, line 34: The reason that I'm using ete3 here instead of inputting the string directly into `populate_tree` is because the formatting is off. Right now, `populate_tree` and `newick_to_network` both assume a specific newick format, specifically those with internal nodes in the newick string. Should I add an argument to `newick_to_network`, `populate_tree`, and the `init` of CassiopeiaTree allowing for the user to specify the newick format?

PR #95, opened 2021-05-20
General bd simulator

richardyz98:
Hi Matt,
This is the PR for the use of the generator in the solver, as well as some clerical fixes.

I added one similarity function with tests. Also, in the NeighborJoiningSolver, we were keeping the root node around after the fact and keeping it in the character matrix as a sample. This potentially throws off downstream stuff so I changed it so the root node no longer persists in the character matrix of the reconstructed tree.

Let me know if you have any questions,
Richard


cassiopeia/solver/SharedMutationJoiningSolver.py
```diff
@@ -84,6 +93,9 @@ def solve(self, cassiopeia_tree: CassiopeiaTree) -> None:
         Args:
             cassiopeia_tree: CassiopeiaTree object to be populated
         """
+
+        names = solver_utilities.node_name_generator()
```
mattjones315, line 97: ditto re: informative name for node name generator

cassiopeia/solver/solver_utilities.py
```diff
@@ -18,6 +18,14 @@ class PriorTransformationError(Exception):
...
 
 
+def node_name_generator() -> Generator[str, None, None]:
+    """Generates unique node names for building the reconstructed tree."""
+    i = 0
+    while True:
+        yield "cassiopeia_internal_node_" + str(i)
+        i += 1
+
+
```
mattjones315, line 28: Some questions on behavior -- 

(1) Every time you call this function does this create a _new_ generator? As in there won't be contamination between generators being used in separate functions right?

(2) What else do you do to ensure that you aren't creating duplicate nodes in the CassiopeiaTree? For example, let's consider a pathological example where a character matrix entries starting with `cassiopeia_internal_node` -- how do we handle this? Is there a more unique identifier that we can use? One possibility is to hash the timestamp and use this as the identifier before the unique integer. Thoughts?

mattjones315 (changes requested):
This is looking nice Richard!

I've left a few comments here -- mostly around the readability of the code. I'm also concerned with guaranteeing that the node names returned by the generator are unique. Can you take a look at my suggestion and see if it make sense?

Finally, I highly suggest that you organize your engineering efforts a bit better. This PR is from the `general_bd_simulator` branch but it has nothing to do with the simulator. Please create new branches for new ideas / work so it's easier to review, organize, and track over time.

Thanks!

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -189,6 +189,10 @@ def __check_network_initialized(self) -> None:
     def set_character_matrix(self, character_matrix: pd.DataFrame):
         """Initializes a character matrix in the object."""
 
+        if not all(type(i) == str for i in character_matrix.index):
```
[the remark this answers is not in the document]
richardyz98: I think it wouldn't work, as it seems the default dtype for a pandas index is `object`

mattjones315 (approved):
All of this looks great! Let's merge it in.

PR #96, opened 2021-05-25
Fix DistanceSolver adding non-unique node names

Lioscro:
I noticed the `solve` method of `DistanceSolver` was using the number of nodes in the current tree as the name of the new node that will be added (the node that connects the two cherries). Initially, the tree is initialized from the nodes in the dissimilarity matrix, which only defines the leaf nodes. The leaf node names in this matrix are not ordered. For example, the following leaves may be present in the dissimilarity matrix: `1`, `3`, `30`, `4`. These nodes are then added into the initial tree. Then, using the number of nodes in the tree (which is 4) as the name of the new node will cause problems. Namely, new edges will be added with leaf `4` as one vertex, instead of a new parent node. Also, this causes duplicate entries in the dissimilarity matrix. This PR updates the new node naming scheme to prefix all new nodes with an underscore to prevent such collisions.

Also, slight improvements to `UPGMASolver` and `NeighborJoiningSolver` and an update in `CassiopeiaTree.collapse_unifurcations` so that the iterator doesn't change during iteration.

### DIFF
--- a/cassiopeia/data/CassiopeiaTree.py
+++ b/cassiopeia/data/CassiopeiaTree.py
@@ -6,8 +6,8 @@ clonal population (though this is not required). Other important data is also
 stored here, like the priors for given character states as well any meta data
 associated with this clonal  population.
 
-When a solver has been called on this object, a tree will be added to the data 
-structure at which point basic properties can be queried like the average tree 
+When a solver has been called on this object, a tree will be added to the data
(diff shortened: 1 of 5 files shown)


Lioscro:
Looks like this is somewhat duplicated to changes made in #95. This PR should be rebased with that once merged.

PR #97, opened 2021-05-25
Supercellular sampler

Lioscro:
`SupercellularSampler` implementation, along with several utility functions in `CassiopeiaTree` that are useful in dealing with ambiguous states and branch distances.

Some important changes are the following.
* Ambiguous characters are denoted as a tuple in the character matrix and/or the network. Updates to some functions and addition of new functions in `CassiopeaTree` to support ambiguous characters.
* When computing dissimilarities with `CassiopeiaTree.compute_dissimilarity_map`, a new `cluster_dissimilarity` function should be used when there are ambiguous characters. Here is an example snippet

```python
from functools import partial
dissimilarity_function = partial(cluster_dissimilarity, weighted_hamming_distance)
tree.compute_dissimilarity_map(dissimilarity_function)
```
I had considered whether this should be done automatically in ``compute_dissimilarity_map``, but felt that the arguments should always be explicit for the user.

* Updated `n_cell`, `n_character` and `populate_tree` in `CassiopeiaTree` to use the current character matrix instead of the original one because these properties/function should always return/operate on the current object state, instead of a possibly-outdated state (i.e. the original character matrix). Previously, the original (old) state was being used and this was causing problems for `GreedySolver` when trying to call `collapse_mutationless_edges`.

This PR is big enough for reviews from both @mattjones315 and @richardyz98.


cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1047,6 +1107,61 @@ def relabel_nodes(self, relabel_map: Dict[str, str]) -> None:
...
+        # Update character matrix, cell meta, disimilarity map if they exist.
+        if self.__current_character_matrix is not None:
+            state = [self.missing_state_indicator] * self.n_character
+            self.__set_character_states(node, state)
+            self.__current_character_matrix.loc[node] = state
+        if self.cell_meta is not None:
+            self.cell_meta.loc[node] = None
+        if self.__dissimilarity_map is not None:
+            self.__dissimilarity_map.loc[node] = np.inf
+            self.__dissimilarity_map[node] = np.inf
```
Lioscro, line 1160: Wait until #95 is resolved, which contains similar operations.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1081,13 +1196,13 @@ def remove_leaf_and_prune_lineage(self, node: str) -> None:
...
-        if self.__current_character_matrix:
+        if self.__current_character_matrix is not None:
             if node in self.__current_character_matrix.index:
                 self.__current_character_matrix.drop(index = [node], inplace = True)
-        if self.cell_meta:
+        if self.cell_meta is not None:
             if node in self.cell_meta.index:
                 self.cell_meta.drop(index = [node], inplace = True)
-        if self.__dissimilarity_map:
+        if self.__dissimilarity_map is not None:
```
Lioscro, line 1205: same here

cassiopeia/simulator/SupercellularSampler.py
```diff
@@ -0,0 +1,126 @@
...
+                leaves, p=np.array(weights) / np.sum(weights)
+            )
+            leaf2_state = merged_tree.get_character_states(leaf2)
+
+            # Merge these two leaves at the mean time of the two leaves
+            # If the tree is ultrametric, this preserves ultrametricity
+            new_leaf = f"{leaf1}-{leaf2}"
+            new_time = (
+                merged_tree.get_time(leaf1) + merged_tree.get_time(leaf2)
+            ) / 2
```
mattjones315, line 104: If the tree isn't ultrametric, is there a chance that this new time will be _before_ the ancestor of the two nodes? If so, what's your fix?

cassiopeia/solver/dissimilarity_functions.py
```diff
@@ -225,3 +221,95 @@ def weighted_hamming_similarity(
...
+        [List[int], List[int], int, Dict[int, Dict[int, float]]], float
+    ],
+    s1: Union[List[int], List[Tuple[int, ...]]],
+    s2: Union[List[int], List[Tuple[int, ...]]],
+    missing_state_indicator: int,
+    weights: Optional[Dict[int, Dict[int, float]]] = None,
+    linkage_function: Callable[[Union[np.array, List[float]]], float] = np.mean,
+    normalize: bool = True,
+) -> float:
+    """This function computes the dissimilarity between two (possibly) ambiguous
```
mattjones315, line 237: ditto: docstring style

cassiopeia/solver/dissimilarity_functions.py
```diff
@@ -225,3 +221,95 @@ def weighted_hamming_similarity(
...
+        present = []
+        for _c1, _c2 in itertools.product(c1, c2):
+            present.append(
+                _c1 != missing_state_indicator
+                and _c2 != missing_state_indicator
+            )
+            dissim.append(
+                dissimilarity_function(
+                    [_c1],
+                    [_c2],
```
mattjones315, line 307: If you're using one of the numba optimized functions, they might start complaining that these aren't numba lists or numpy arrays -- can you wrap these with np.array? Or am I incorrect?

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1334,14 +1568,149 @@ def filter_nodes(self, condition: Callable[[str], bool]) -> List[str]:
...
+                    ancestor_distance = node_time - ancestor_time
+                    distances[ancestor] = ancestor_distance
+
+                    for descendant in self.depth_first_traverse_nodes(
+                        ancestor, postorder=False
+                    ):
+                        if descendant not in distances:
+                            distances[descendant] = ancestor_distance + (
+                                self.get_time(descendant) - ancestor_time
+                            )
```
mattjones315, line 1707: Isn't this code block going to reconsider nodes several times? Why not just do a single DFS from the root, ignoring descendants of the query `node` passed into the function?

cassiopeia/simulator/SupercellularSampler.py
```diff
@@ -0,0 +1,126 @@
...
+            merged_tree.set_character_states(new_leaf, new_state)
+            merged_tree.remove_leaf_and_prune_lineage(leaf1)
+            merged_tree.remove_leaf_and_prune_lineage(leaf2)
+
+        if collapse_source is None:
+            collapse_source = merged_tree.root
+        merged_tree.collapse_unifurcations(source=collapse_source)
+
+        if collapse_duplicates:
+            merged_tree.collapse_ambiguous_characters()
```
mattjones315, line 124: Is there a use case where you would not want to collapse ambiguous characters?

mattjones315 (changes requested):
This is a fantastic PR! Really impressive work. Not quite ready to merge in yet though, there are a few aesthetic things that I'd like to clean up to make it more aligned with the rest of the codebase's style (mainly around docstrings) and I have some questions that I'd like to resolve before giving approval. @richardyz98 should also take a pass here to review this, as I'm sure I've missed some things. Of course, happy to take another look when some of these issues are resolved.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -162,12 +163,12 @@ def populate_tree(self, tree: Union[str, ete3.Tree, nx.DiGraph]) -> None:
...
-                and n in self.__original_character_matrix.index.tolist()
+                self.__current_character_matrix is not None
+                and n in self.__current_character_matrix.index.tolist()
             ):
                 self.__network.nodes[n][
                     "character_states"
-                ] = self.__original_character_matrix.loc[n].to_list()
+                ] = self.__current_character_matrix.loc[n].to_list()
             else:
                 self.__network.nodes[n]["character_states"] = []
```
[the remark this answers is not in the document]
Lioscro: The reason is twofold: 1) exactly what you mentioned about the sampling procedure. One thing that you may be mistaken is that the `SupercellularSampler` retains the original character matrix (with unambiguous cellular leaf names and states) and only modifies the current character matrix to contain merged states. When `populate_tree` is called on this tree, it tries to look up the old, cellular leaf states, which are now outdated.
I guess we could circumvent this by just setting the original & current character matrix to be the merged one, but either way I still think every operation should be done on the *current* object state rather than a (possibly) outdated original character matrix. My understanding of the `__original_character_matrix` is that it is like `adata.raw` where it exists if the user wants to recover the original version, but should not be used when operating on the anndata itself.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -469,13 +470,101 @@ def is_internal_node(self, node: str) -> bool:
         self.__check_network_initialized()
         return self.__network.out_degree(node) > 0
 
+    def is_ambiguous(self, node: str) -> bool:
+        """Returns whether the node is ambiguous. This is detected by checking if
+        any of the characters are tuples.
+
```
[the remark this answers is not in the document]
Lioscro: Thanks, will update these.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -469,13 +470,101 @@ def is_internal_node(self, node: str) -> bool:
...
+        Raises:
+            CassiopeiaTreeError if the tree has not been initialized.
+        """
+        self.__check_network_initialized()
+        states = self.get_character_states(node)
+        return any(isinstance(state, tuple) for state in states)
+
+    def collapse_ambiguous_characters(self) -> None:
+        """Only retain unique characters for ambiguous nodes. Usually, ambiguous
+        nodes have character strings represented as a list of tuples of integers.
```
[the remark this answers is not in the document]
Lioscro: This is a typo. Ambiguous nodes should always have list of tuples as their character string.

The rationale is that we can work with the relative abundances (think number of UMIs), and it is possible to recover the original distribution of the indel states. This is actually what I'm testing in my simulations to see if it actually makes a difference.

I guess the short version is that we can go from UMI table -> character matrix in a non-destructive way, while also allowing it to be collapsed if wanted.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -848,11 +946,15 @@ def get_all_ancestors(self, node: str) -> List[str]:
...
-            self.__cache["ancestors"][node] = [n
-                for n in nx.ancestors(self.__network, node)
-            ]
+            self.__cache["ancestors"][node] = []
+
+            current_node = node
+            while not self.is_root(current_node):
+                ancestor = self.parent(current_node)
+                self.__cache["ancestors"][node].append(ancestor)
+                current_node = ancestor
```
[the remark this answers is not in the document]
Lioscro: Don't know about speed, but it guarantees the node ordering. The previous implementation returned a random order.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1053,28 +1163,99 @@ def relabel_nodes(self, relabel_map: Dict[str, str]) -> None:
...
+        (if there is a character matrix), an empty cell meta (if it exists), and
+        infinite dissimlarity with all other leaves. The current tree is modified
+        in-place.
+
+        The *caller* is expected to maintain any desired constraints (such as ultrametricity)
+        by calling :func:`set_time` and/or :func:`set_branch_length`, as well as
+        character states by calling :func:`set_character_states`. If this function
+        is called on a tree with an existing dissimilarity map, it is expected the
+        caller will re-run dissimilarity calculation, because the new leaf is
+        initialized to have infinite dissimilarity with all other leaves.
```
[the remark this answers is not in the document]
Lioscro: Sure

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1127,7 +1308,7 @@ def collapse_unifurcations(self, source: Optional[int] = None) -> None:
         if source is None:
             source = self.root
 
-        for node in self.depth_first_traverse_nodes(postorder = True, source = source):
+        for node in list(self.depth_first_traverse_nodes(postorder = True, source = source)):
```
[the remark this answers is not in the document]
Lioscro: Was getting `RuntimeError: dictionary keys changed during iteration`

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1179,7 +1360,7 @@ def collapse_mutationless_edges(
         if infer_ancestral_characters:
             self.reconstruct_ancestral_characters()
 
-        for n in self.depth_first_traverse_nodes(postorder = True):
+        for n in list(self.depth_first_traverse_nodes(postorder = True)):
```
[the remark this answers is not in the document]
Lioscro: same as above

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1276,6 +1475,7 @@ def compute_dissimilarity_map(
             dissimilarity_function,
             weights,
             self.missing_state_indicator,
+            numbaize=numbaize
```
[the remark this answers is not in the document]
Lioscro: Unfortunately, I've spent way too much time trying to get it to numbaize with no progress. Numba doesn't play well with variable length anything, and since each tuple may have variable number of states this isn't easily numbaizable :(

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1334,14 +1534,119 @@ def filter_nodes(self, condition: Callable[[str], bool]) -> List[str]:
...
+
+        Raises:
+            CassiopeiaTreeError if the tree has not been initialized.
         """
         self.__check_network_initialized()
         return nx.tree_all_pairs_lowest_common_ancestor(self.__network, root=self.root, pairs=pairs)
+
+    def find_lca(self, *nodes: str) -> str:
+        """Finds the LCA of all provided nodes. Internally, this function iteratively
+        calls :func:`find_lcas_of_pairs`.
```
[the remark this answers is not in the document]
Lioscro: Good idea.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1334,14 +1534,119 @@ def filter_nodes(self, condition: Callable[[str], bool]) -> List[str]:
...
+        if "distances" not in self.__cache \
+            or node1 not in self.__cache["distances"] \
+            or node2 not in self.__cache["distances"][node1]:
+            distances = self.__cache.setdefault("distances", {}).setdefault(node1, {})
+            if node1 == node2:
+                distances[node2] = 0
+            else:
+                lca = self.find_lca(node1, node2)
+                lca_time = self.get_time(lca)
+                distances[node2] = (self.get_time(node1) - lca_time) + (self.get_time(node2) - lca_time)
```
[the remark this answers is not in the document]
Lioscro: Don't think its worth caching this as long as we cache all ancestors for each node.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1334,14 +1534,119 @@ def filter_nodes(self, condition: Callable[[str], bool]) -> List[str]:
...
+        it uses properties of trees to compute these much quicker, and the results
+        are cached.
+
+        Args:
+            node: Node from which to compute distance to all other nodes
+            leaves_only: Calculate distances to leaves only and not internal nodes.
+                Defaults to False.
+
+        Returns:
+            Dictionary of distances
```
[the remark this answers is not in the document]
Lioscro: I think dictionary of distances work fine here. A dataframe might be suitable when we want all pairwise distances (but this function only returns distances from a single node to all other nodes).

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1334,14 +1534,119 @@ def filter_nodes(self, condition: Callable[[str], bool]) -> List[str]:
...
+
+        Raises:
+            CassiopeiaTreeError if the tree has not been initialized.
+        """
+        self.__check_network_initialized()
+
+        if "distances" not in self.__cache or node not in self.__cache["distances"] or any(
+            _node not in self.__cache["distances"][node] for _node in self.nodes
+        ):
+            distances = self.__cache.setdefault("distances", {}).setdefault(node, {})
```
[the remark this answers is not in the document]
Lioscro: I updated this function to use `nx.shortest_path_length` instead, which can be configured (depending on what optional parameters are provided) to compute all distances from a single node to all other nodes.

I think in this case computing only the distances we want separately may give us worse performance due to the many `nx.shortest_path_length` that would be required.

cassiopeia/simulator/SupercellularSampler.py
Lioscro: Yes, if you want to weigh each state by relative abundance.

cassiopeia/data/CassiopeiaTree.py
Lioscro: Changed this to use `nx. shortest_path_length` instead.

cassiopeia/solver/dissimilarity_functions.py
Lioscro: The `dissimilarity_function` is not numbaized because the overhead due to calling the numbaized function multiple times (with very small arguments) overweighs any speedup (around 100x slower).

cassiopeia/simulator/SupercellularSampler.py
Lioscro: Good point. I'll choose the average time or the time of the LCA, whichever is greater.

Lioscro: Hold up, since we are adding the leaf to the LCA, which MUST have time less than or equal to the two chosen leaves. Therefore, this can never happen.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -469,13 +470,101 @@ def is_internal_node(self, node: str) -> bool:
...
+                if isinstance(states[i], tuple):
+                    new_states = tuple(set(states[i]))
+                    if states != new_states:
+                        states[i] = new_states
+                        modified = True
+            if modified:
+                self.set_character_states(node, states)
+
+    def resolve_ambiguous_characters(
+        self, resolve_function: Optional[Callable[[List[int]], int]] = None
```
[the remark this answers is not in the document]
Lioscro: Done!

cassiopeia/data/utilities.py
```diff
@@ -144,12 +145,24 @@ def compute_dissimilarity_map(
...
     Returns:
         A dissimilarity mapping as a flattened array.
     """
 
-    nb_dissimilarity = numba.jit(dissimilarity_function, nopython=True)
+    # Try to numbaize the dissimilarity function, but fallback to python
+    try:
+        nopython = True
+        dissimilarity_func = numba.jit(dissimilarity_function, nopython=True)
+    except Exception:
```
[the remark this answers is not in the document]
Lioscro: Done.

cassiopeia/data/CassiopeiaTree.py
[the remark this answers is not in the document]
mattjones315: I see - this makes sense!

I think that this is potentially another reason to create several "layers" for character matrices like in Scanpy. That way you can keep around this full character matrix should you wish to.

cassiopeia/solver/dissimilarity_functions.py
mattjones315: Perhaps I'm misunderstanding something, but isn't that precisely when Numba should be given us a speedup? I understand that if you need to recompile a Numba function several times it might slow us down, but if you only compile once then the function should speed things up if you call it several times, right?

mattjones315 (changes requested):
Still looking great! I left a few comments I'd like to get ironed out before merging this in. Thanks!

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -803,9 +900,21 @@ def set_character_states(self, node: str, states: List[int]) -> None:
         self.__set_character_states(node, states)
 
         if self.is_leaf(node):
-            self.__current_character_matrix.loc[node] = states
-
-    def __set_character_states(self, node: str, states: List[int]) -> None:
+            # Cast entire character matrix to object dtype if we get an ambiguous
+            # character string.
+            if any(isinstance(state, tuple) for state in states):
```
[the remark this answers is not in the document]
Lioscro: Good point. For some reason I thought the character states haven't been updated in the network yet (but it actually was with the previous `__set_character_states` call.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1179,13 +1429,15 @@ def collapse_mutationless_edges(
         if infer_ancestral_characters:
             self.reconstruct_ancestral_characters()
 
-        for n in self.depth_first_traverse_nodes(postorder = True):
+        for n in list(self.depth_first_traverse_nodes(postorder=True)):
```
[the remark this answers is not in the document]
Lioscro: This is similar to your previous comment. I kept getting `RuntimeError: dictionary key changed during iteration`.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1228,6 +1480,60 @@ def set_dissimilarity_map(self, dissimilarity_map: pd.DataFrame) -> None:
...
+        """
+        if self.__dissimilarity_map is None:
+            raise CassiopeiaTreeError("No dissimilarity map is defined.")
+        if node not in self.__dissimilarity_map.index:
+            raise CassiopeiaTreeError(
+                f"`{node}` is not in the dissimilarity map."
+            )
+        return self.__dissimilarity_map.loc[node].to_dict()
+
+    def set_dissimilarity(
```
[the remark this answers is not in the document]
Lioscro: For example, when a new leaf is added and we want to (partially) update the dissimilarity map, instead of re-computing the entire thing. In this case, because `add_leaf` calls `__register_data_with_tree`, which also adds the leaf (with infinite dissimilarity to all other nodes), we need to be able to overwrite existing dissimilarities to be able to set the correct values for this leaf after calling `add_leaf`.

cassiopeia/data/utilities.py
```diff
@@ -1,7 +1,9 @@
 """
 General utilities for the datasets encountered in Cassiopeia.
 """
-from typing import Callable, Dict, List, Optional, Tuple
+import collections
+import warnings
+from typing import Callable, Dict, List, Optional, Tuple, Union
```
mattjones315, line 6: We introduce imports alphabetically - can you move `typing` above `warnings`?

cassiopeia/data/utilities.py
```diff
@@ -36,7 +45,12 @@ def get_lca_characters(
         assert len(i) == k
     lca_vec = [0] * len(vecs[0])
     for i in range(k):
-        chars = set([vec[i] for vec in vecs])
+        chars = set()
+        for vec in vecs:
+            if isinstance(vec[i], tuple):
```
mattjones315, line 50: why not use `is_ambiguous_state`, since you defined it at the bottom?

mattjones315 (changes requested):
Just one or two more comments to address, else I think we're ready to merge in!

cassiopeia/data/CassiopeiaTree.py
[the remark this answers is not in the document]
Lioscro: Agreed

cassiopeia/data/utilities.py
Lioscro: Ha. Seems like I forgot my abc's :sweat_smile:

cassiopeia/data/utilities.py
Lioscro: Yep, good catch

Lioscro:
Thanks for the helpful reviews @mattjones315! Almost ready for merge. :)

mattjones315 (approved):
This looks great to me! Thanks for the hard work here, I think we're ready to merge it in.

Lioscro:
@mattjones315 and I went through some extensive reviews, but could you also take a look @richardyz98? Would appreciate any feedback!

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1334,14 +1651,130 @@ def filter_nodes(self, condition: Callable[[str], bool]) -> List[str]:
...
+            frozenset({node, _node}) not in self.__cache["distances"]
+            for _node in self.nodes
+        ):
+            distances = self.__cache.setdefault("distances", {})
+            undirected_network = self.__network.to_undirected()
+            for other_node, distance in nx.shortest_path_length(
+                undirected_network, source=node, weight="time"
+            ).items():
+                distances[frozenset({node, other_node})] = distance
+
```
richardyz98, line 1776: Should the weight not be "length" if we are to be weighting over branch lengths? "time" refers to the node time.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1053,28 +1193,141 @@ def relabel_nodes(self, relabel_map: Dict[str, str]) -> None:
...
+            add_to_character_matrix = leaves_set - set(
+                self.__current_character_matrix.index
+            )
+            for to_add in add_to_character_matrix:
+                # Note that we initialize state each iteration so that each state
+                # is a new object.
+                state = [self.missing_state_indicator] * self.n_character
+                self.__set_character_states(to_add, state)
+                self.__current_character_matrix.loc[to_add] = state
```
richardyz98, line 1224: I think perhaps it'd be more clear to refer to it as a "character vector" rather than "state", as "state" already carries the meaning of an indel

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1334,14 +1651,130 @@ def filter_nodes(self, condition: Callable[[str], bool]) -> List[str]:
...
+        key = frozenset({node1, node2})
+        if (
+            "distances" not in self.__cache
+            or key not in self.__cache["distances"]
+        ):
+            distances = self.__cache.setdefault("distances", {})
+            undirected_network = self.__network.to_undirected()
+            distances[key] = nx.shortest_path_length(
+                undirected_network, source=node1, target=node2, weight="time"
+            )
```
richardyz98, line 1741: "time" is used for nodes, the attribute for branch length is "length". I tested it really quickly and this just returns the unweighted path length (i.e. number of seperating) for nodes. Key to fix this or else the sampling in the merging step will not be correct!

richardyz98 (changes requested):
Hey Joseph,
Honestly amazing PR. I'm impressed by how quickly you found your way around the codebase! 

I had some minor requests regarding the naming conventions around "states". Also there is one major issue with `get_distance`, please check that.

Thanks for the amazing work!

cassiopeia/data/CassiopeiaTree.py
Lioscro: Oof. Thanks for catching this!
Thankfully for my simulations I used an older implementation of this where I manually calculate the branch distance (instead of using networkx), so they should be fine.

cassiopeia/solver/NeighborJoiningSolver.py
```diff
@@ -255,8 +256,25 @@ def setup_root_finder(self, cassiopeia_tree: CassiopeiaTree) -> None:
...
+                    weights = solver_utilities.transform_priors(
+                        cassiopeia_tree.priors, self.prior_transformation
+                    )
+                dissimilarity[leaf] = self.dissimilarity_function(
+                    rooted_character_matrix.loc["root"].values,
+                    rooted_character_matrix.loc[leaf].values,
+                    cassiopeia_tree.missing_state_indicator,
+                    weights,
+                )
+            cassiopeia_tree.set_dissimilarity("root", dissimilarity)
```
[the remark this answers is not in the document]
Lioscro: I think you're missing the if statement. Previously, we would always re-calculate the entire dissimilarity map, regardless of whether or not the tree already had one calculated. So what I'm doing here is, if the dissimilarity map is already defined (i.e. calculated) for the tree, only computing & adding the dissimilarities from the `"root"` to all other leaves. If the dissimilarity map is not defined, the entire map is computed (same behavior as before). Does that make sense?

richardyz98 (approved):

PR #98, opened 2021-06-04
ILPSolver Potential Graph Speedup & Refactor

mattjones315:
Provided new engineering on the backend of the `ILPSolver` to speed up the potential graph inference. This included moving some functions previously in the `ILPSolver` python file to the `ilp_solver_utilities.pyx` cython file, and refactoring the data structures used during potential graph inference.

### DIFF
--- a/cassiopeia/solver/ILPSolver.py
+++ b/cassiopeia/solver/ILPSolver.py
@@ -50,6 +50,10 @@ class ILPSolver(CassiopeiaSolver.CassiopeiaSolver):
         maximum_potential_graph_layer_size: Maximum size allowed for an iteration
             of the potential graph inference procedure. If this is exceeded,
             we return the previous iteration's graph or abort altogether.
+        maximum_potential_graph_lca_height: Maximum height of LCA to add to the
+            potential graph. If this parameter is not provided or the specified
+            value is 0, the maximum distance between any pair of samples is used
+            as the maximum lca height.
         weighted: Weight edges on the potential graph by the negative log
             likelihood of the mutations.
         seed: Random seed to use during ILP optimization.
@@ -71,6 +75,7 @@ class ILPSolver(CassiopeiaSolver.CassiopeiaSolver):
         convergence_time_limit: int = 12600,
         convergence_iteration_limit: int = 0,
         maximum_potential_graph_layer_size: int = 10000,
+        maximum_potential_graph_lca_height: Optional[int] = None,
         weighted: bool = False,
         seed: Optional[int] = None,
         mip_gap: float = 0.01,
@@ -82,6 +87,9 @@ class ILPSolver(CassiopeiaSolver.CassiopeiaSolver):
         self.convergence_iteration_limit = convergence_iteration_limit
         self.maximum_potential_graph_layer_size = (
             maximum_potential_graph_layer_size
+        )
+        self.maximum_potential_graph_lca_height = (
+            maximum_potential_graph_lca_height
         )
         self.weighted = weighted
         self.seed = seed
@@ -142,21 +150,25 @@ class ILPSolver(CassiopeiaSolver.CassiopeiaSolver):
 
         # determine diameter of the dataset by evaluating maximum distance to
         # the root from each sample
-        max_lca_distance = 0
-        lca_distances = [
-            dissimilarity_functions.hamming_distance(
-                root,
-                np.array(u),
-                ignore_missing_state=True,
-                missing_state_indicator=cassiopeia_tree.missing_state_indicator,
-            )
-            for u in targets
-        ]
+        if (self.maximum_potential_graph_lca_height is not None) and (self.maximum_potential_graph_lca_height > 0):
+            max_lca_distance = self.maximum_potential_graph_lca_height
+        
+        else:
+            max_lca_distance = 0
+            lca_distances = [
+                dissimilarity_functions.hamming_distance(
+                    root,
+                    np.array(u),
+                    ignore_missing_state=True,
+                    missing_state_indicator=cassiopeia_tree.missing_state_indicator,
+                )
+                for u in targets
+            ]
 
-        for (i, j) in itertools.combinations(range(len(lca_distances)), 2):
-            max_lca_distance = max(
-                max_lca_distance, lca_distances[i] + lca_distances[j] + 1
-            )
+            for (i, j) in itertools.combinations(range(len(lca_distances)), 2):
+                max_lca_distance = max(
+                    max_lca_distance, lca_distances[i] + lca_distances[j] + 1
+                )
 
         # infer the potential graph
         potential_graph = self.infer_potential_graph(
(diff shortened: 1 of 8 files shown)


cassiopeia/solver/ilp_solver_utilities.pyx
```diff
@@ -1,22 +1,157 @@
...
+                for e in layer_edges
+                if "|".join(e.split("|")[:n_characters])
+                != "|".join(e.split("|")[n_characters:])
+            ]
+            current_layer_edges += layer_edges
+
+            if source_nodes.shape[0] > next_layer.shape[0]:
+                if effective_threshold == distance_threshold:
+                    effective_threshold *= 3
+            source_nodes = next_layer
```
richardyz98, line 131: remind me, is there a reason behind the 3? Or is it just arbitrary

cassiopeia/solver/ilp_solver_utilities.pyx
```diff
@@ -93,13 +237,90 @@ def infer_layer_of_potential_graph(
...
+                ancestor[i] = arr2[i]
+            elif arr2[i] == missing_state_indicator:
+                ancestor[i] = arr1[i]
+
+    return "|".join(ancestor)
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def simple_hamming_distance_cython(
```
richardyz98, line 299: Hey Matt so I'm not super familiar with cython, do we need both this cythonized version and the non-cythonized version of this function because we have to call something like this in cythonized functions and non-cythonized functions? We can't collapse these?

richardyz98 (changes requested):
Hey Matt,
Thanks so much for handling the cythonizing of ILP. It looks very clean, and I don't have many comments, just a few questions mostly.

One thing that I have noted is that in order to calculate LCA distance, we use a hamming distance (ignoring missing). But for the calculation of the actual edge weights in the potential graph, we used our bespoke weighted hamming distance. Is this because in calculating the LCA distance, we're just trying to limit the diameter of the potential graph in # of mutations so we don't need to be as exact?

Thanks for answering my inquiries!

cassiopeia/solver/ilp_solver_utilities.pyx
```diff
@@ -1,22 +1,157 @@
...
+
+            (next_layer, layer_edges) = infer_layer_of_potential_graph(
+                source_nodes, effective_threshold
+            )
+
+            # subset to unique values
+            if next_layer.shape[0] > 0:
+                next_layer = np.unique(next_layer)
+                # unique_idx = fast_unique(np.array([node.split("|") for node in next_layer]).astype(int))
+                # next_layer = next_layer[unique_idx]
```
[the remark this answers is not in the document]
mattjones315: No reason to keep it -- I just had kept it around before I was sure I wanted to get rid of it.

test/solver_tests/ilp_solver_test.py
```diff
@@ -6,6 +6,7 @@
 
 import itertools
 import networkx as nx
+from networkx.algorithms.operators.binary import intersection
```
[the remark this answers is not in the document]
mattjones315: good call.

cassiopeia/solver/ilp_solver_utilities.pyx
mattjones315: No reason -- this was just in the original codebase that Alex and I developed a few years ago. I think Alex chose this modifier because it worked well empirically

cassiopeia/solver/ilp_solver_utilities.pyx
mattjones315: Two reason -- the first is that the cythonized version works a lot faster here, and it operates on strings not integers.

Also, I'm intentionally ignoring missing data here, which is how the original codebase worked (I think it's okay for inferring the potential graph).

cassiopeia/solver/ILPSolver.py
```diff
@@ -142,16 +150,25 @@ def solve(
 
         # determine diameter of the dataset by evaluating maximum distance to
         # the root from each sample
-        max_lca_distance = 0
-        lca_distances = [
-            dissimilarity_functions.hamming_distance(root, np.array(u))
-            for u in targets
-        ]
+        if (self.maximum_potential_graph_lca_height is not None) and (self.maximum_potential_graph_lca_height > 0):
+            max_lca_distance = self.maximum_potential_graph_lca_height
```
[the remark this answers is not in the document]
mattjones315: No, I don't think it will because of the `else` clause right below. We only calculate the `maximum_potential_graph_lca_height` is not specified.

This also seems to work empirically when I run the code.

mattjones315:
Thanks for the feedback Richard!

I've made a few changes and have formatted the code a bit more. Can you let me know if there's anything else you'd like to discuss / see changed? Also, can you make sure that the code runs alright on your end (specifically, all tests pass)?

cassiopeia/solver/ilp_solver_utilities.pyx
```diff
@@ -1,22 +1,151 @@
...
-import pandas as pd
 
 from cassiopeia.data import utilities as data_utilities
 from cassiopeia.solver import dissimilarity_functions
 
 
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def infer_potential_graph_cython(
+    character_array: str[:, :],
```
Lioscro, line 21: What's the rationale for using strings to represent characters instead of integers, like we normally do? Could we add documentation somewhere why strings are better?

cassiopeia/solver/ilp_solver_utilities.pyx
Lioscro: Should we be converting all of our dissimilarity functions to cython at some point?

test/solver_tests/dissimilarity_functions_test.py
```diff
@@ -253,6 +253,21 @@ def test_weighted_hamming_similarity_all_missing(self):
...
+    def test_hamming_distance_ignore_missing(self):
+
+        distance = dissimilarity_functions.hamming_distance(self.s1, self.s2, ignore_missing_state=True)
+
+        self.assertEqual(distance, 2)
+
+        distance = dissimilarity_functions.hamming_distance(self.s1, self.all_missing, ignore_missing_state=True)
+
+        self.assertEqual(distance, 0)
+
```
Lioscro, line 271: Pretty long line here. Not sure if we're supposed to run `black` on tests, but if we do, this should be broken up by running it. Otherwise, feel free to resolve and ignore.

test/solver_tests/ilp_solver_test.py
```diff
@@ -91,6 +91,30 @@ def setUp(self):
...
+    def test_get_lca_cython(self):
+
+        # test single sample
+        cm = self.missing_tree.get_current_character_matrix().astype(str)
+
+        lca = ilp_solver_utilities.get_lca_characters_cython(cm.loc["a"].values, cm.loc["b"].values, 4, "-1")
+        
+        self.assertEqual(lca, "1|3|1|1")
+
+        lca = ilp_solver_utilities.get_lca_characters_cython(cm.loc["h"].values, cm.loc["b"].values, 4, "-1")
```
Lioscro, line 103: ditto

test/solver_tests/ilp_solver_test.py
```diff
@@ -91,6 +91,30 @@ def setUp(self):
...
+    def test_cython_hamming_dist(self):
+
+        sample1 = np.array(['1', '2', '3', '0', '0'])
+        sample2 = np.array(['1', '4', '0', '0', '1'])
+        dist = ilp_solver_utilities.simple_hamming_distance_cython(sample1, sample2, '-')
+        self.assertEqual(dist, 3)
+
+        sample1 = np.array(['1', '2', '3', '0', '-'])
+        sample2 = np.array(['1', '-', '0', '0', '1'])
+        dist = ilp_solver_utilities.simple_hamming_distance_cython(sample1, sample2, '-')
```
Lioscro, line 115: ditto

test/solver_tests/ilp_solver_test.py
```diff
@@ -142,15 +166,17 @@ def test_get_layer_for_potential_graph(self):
         source_nodes = unique_character_matrix.values
         dim = source_nodes.shape[1]
 
+        source_node_strings = np.array(["|".join(arr) for arr in source_nodes.astype(str)])
```
Lioscro, line 169: ditto

Lioscro (changes requested):
Few minor comments and clarification questions. How much do these changes improve the runtime?

Also, I've confirmed everything compiles and installs fine on my end. All tests passing.

cassiopeia/solver/ilp_solver_utilities.pyx
mattjones315: Though I originally was using integers like elsewhere, I noticed that numpy arrays of integers (and strings for that matter) take up a lot more space than simple strings. So, here I'm essentially compressing character states into character strings of the form `"char1|char2|..."` for memory efficiency. I'll make a note of this in the documentation.

cassiopeia/solver/ilp_solver_utilities.pyx
```diff
@@ -93,13 +231,90 @@ def infer_layer_of_potential_graph(
...
+                ancestor[i] = arr2[i]
+            elif arr2[i] == missing_state_indicator:
+                ancestor[i] = arr1[i]
+
+    return "|".join(ancestor)
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def simple_hamming_distance_cython(
```
richardyz98, line 293: Given that this is ignoring missing data should we add this in the function name?

test/solver_tests/ilp_solver_test.py
```diff
@@ -159,6 +197,9 @@ def test_get_layer_for_potential_graph(self):
             self.assertIn(sample, layer_nodes)
 
         # layer_edges = [(list(e[0]), list(e[1])) for e in layer_edges]
```
richardyz98, line 199: We should drop this?

test/solver_tests/ilp_solver_test.py
```diff
@@ -196,7 +237,7 @@ def test_simple_potential_graph_inference(self):
         )
         max_lca_height = 10
         potential_graph = self.ilp_solver.infer_potential_graph(
-            unique_character_matrix,
+            unique_character_matrix.astype(str),
```
richardyz98, line 240: Is this casting necessary? Doesn't this function perform this cast already before passing on to the cythonized function?

test/solver_tests/ilp_solver_test.py
```diff
@@ -308,7 +349,7 @@ def test_potential_graph_inference_with_duplicates(self):
...
```
richardyz98, line 352: ditto as above

richardyz98 (changes requested):

cassiopeia/solver/dissimilarity_functions.py
```diff
@@ -161,14 +161,22 @@ def hamming_similarity_normalized_over_missing(
...
-def hamming_distance(s1: np.array(int), s2: np.array(int)) -> int:
+def hamming_distance(
+    s1: np.array(int),
+    s2: np.array(int),
+    ignore_missing_state: bool = False,
+    missing_state_indicator: int = -1,
+) -> int:
     """Computes the vanilla hamming distance between two samples.
 
     Counts the number of positions that two samples disagree at.
```
richardyz98, line 172: Should we add to the name and docstring that this ignores missing data?

richardyz98 (changes requested):

cassiopeia/solver/dissimilarity_functions.py
mattjones315: But it doesn't ignore missing data all the time, so I don't want to add it to the function name.

I've added some more details to the docstring.

cassiopeia/solver/ilp_solver_utilities.pyx
mattjones315: I think it's fine -- it shouldn't be exposed to users and I believe this information is in the docstring.

test/solver_tests/dissimilarity_functions_test.py
mattjones315: I've run black on this file now.

test/solver_tests/ilp_solver_test.py
mattjones315: Good point.

test/solver_tests/ilp_solver_test.py
mattjones315: You're right, it's not completely necessary. Thanks!

mattjones315 (commented):
Thanks for the feedback everyone! I've made some changes as you all have suggested and am ready for another review.

richardyz98 (approved):
Looks Great Matt! Thanks again for tackling this!

cassiopeia/solver/ilp_solver_utilities.pyx
Lioscro: Yeah, this probably requires a bit more discussion. I can definitely see the benefit of using cython over numba though -- no need to go through the process of numbaizing the dissimilarity functions and we can probably cythonize the cluster dissimilarity function as well.
Though for most use cases cython is a bit slower than numba 😢

Lioscro (approved):
lgtm

PR #100, opened 2021-06-18
Complete binary tree simulator

Lioscro:
Implementation of `CompleteBinarySimulator`, a subclass of `TreeSimulator` that simulates complete binary trees. Original code from @richardyz98.


mattjones315 (approved):
Looks good to me!

PR #104, opened 2021-07-10
intbc correction to whitelist in preprocessing pipeline

Lioscro:
This PR implements intBC correction to a whitelist.
It uses Levenshtein distance, as implemented in `ngs_tools`, instead of the `Levenshtein` library. This is so that ambiguous bases are treated appropriately. For instance, the `Levenshtein` library has no notion of ambiguous bases, so `ATCN` and `ATCC` has distance of 1. However, the implementation in `ngs_tools` treats `N` and `C` as identical, so distance is 0.

### DIFF
--- a/cassiopeia/preprocess/__init__.py
+++ b/cassiopeia/preprocess/__init__.py
@@ -6,6 +6,7 @@ from .pipeline import (
     align_sequences,
     call_alleles,
     collapse_umis,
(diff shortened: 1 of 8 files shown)


Lioscro:
Closing so that everything is merged into preprocess-ambiguous so that I can test everything at once.

PR #105, opened 2021-07-12
bayesian consensus calling implementation

Lioscro:
Added a `method` keyword argument to `pp.collapse_umis`, which can be specified as either `cutoff` (the old implementation) or `bayesian` (the new implementation).


cassiopeia/preprocess/UMI_utils.py
```diff
@@ -5,20 +5,21 @@
...
 from hits import fastq, utilities, sw, sam
 import logging
+import ngs_tools as ngs
 import numpy as np
 import pandas as pd
 from pathlib import Path
 import pysam
-
 from tqdm.auto import tqdm
+from typing_extensions import Literal
```
mattjones315, line 22: Let's move this up to be right below the first imports from `typing`

cassiopeia/preprocess/UMI_utils.py
```diff
@@ -68,6 +69,10 @@
 filter_func = lambda al: al.has_tag(CELL_BC_TAG)
 
 
+class UMIUtilsError(Exception):
+    pass
+
+
```
mattjones315, line 75: I'd opt for using a general `PreprocessingError` -- thoughts?

cassiopeia/preprocess/UMI_utils.py
```diff
@@ -215,20 +231,28 @@ def form_collapsed_clusters(
...
+                    clusters = form_clusters(
+                        UMI_group, max_read_length, max_hq_mismatches
+                    )
+                elif method == "bayesian":
+                    clusters = form_clusters_bayesian(
+                        UMI_group,
+                        proportion=max_hq_mismatches / max_read_length,
+                    )
+                else:
+                    raise UMIUtilsError(f"Unknown method `{method}`")
```
mattjones315, line 255: Let's make this error message a bit more specific -- something like `Unknown method in forming UMI clusters: {method}`

cassiopeia/preprocess/UMI_utils.py
```diff
@@ -215,20 +231,28 @@ def form_collapsed_clusters(
...
-                    UMI_group, max_read_length, max_hq_mismatches
-                )
+                if method == "cutoff":
+                    clusters = form_clusters(
+                        UMI_group, max_read_length, max_hq_mismatches
+                    )
+                elif method == "bayesian":
+                    clusters = form_clusters_bayesian(
+                        UMI_group,
+                        proportion=max_hq_mismatches / max_read_length,
```
mattjones315, line 252: Do we have any worry that `max_hq_mismatches > max_read_length`?

cassiopeia/preprocess/UMI_utils.py
```diff
@@ -322,6 +346,64 @@ def form_clusters(
...
+    """Forms clusters from a list of aligned segments (reads) probabilistically.
+
+    This function has the same purpose as :func:`form_clusters`, but instead
+    of using hardcoded quality thresholds, takes a probabilisitc approach to
+    forming read clusters by treating the quality scores as probabilities (in
+    fact, they are actually encoded probabilities of error). The two optional
+    arguments specify a dynamic threshold to assign a specific sequence to
+    a cluster. Specifically, for a given cluster sequence, a sequence is
+    assigned to this cluster (consensus) if its "match probability" to this
+    consensus is <= max(min(match probability), `q_threshold` * (`proportion` * length of longest sequence))
```
mattjones315, line 363: this line is too long

cassiopeia/preprocess/UMI_utils.py
```diff
@@ -322,6 +346,64 @@ def form_clusters(
...
+    The "match probability" of a sequence to a possible consensus is is the sum
+    of the quality values where they do not match (equivalent to negative log
+    probability that all mismatches were sequencing errors).
```
mattjones315, line 366: It seems that the match probability is defined as a sum of log probabilities but the thresholds are defined as probabilities (e.g., `q_threshold * proportion * length`). It sounds like you should be taking the log of this threshold probability that way the match probabilities can be compared to your threshold. Are you doing this in `ngs`?

cassiopeia/preprocess/UMI_utils.py
```diff
@@ -322,6 +346,64 @@ def form_clusters(
     return clusters
 
 
+def form_clusters_bayesian(
```
mattjones315, line 349: A note on naming -- is this actually Bayesian? Is this match probability a posterior P(match | data) and you're using this to classify reads? I suppose this could as easily be called `form_clusters_probabilistic`, unless there's a clear bayesian decision rule that I'm missing.

mattjones315 (changes requested):
This is looking great! I have a few questions and comments I'd like to get resolved before merging in. Thanks in advance!

cassiopeia/preprocess/UMI_utils.py
Lioscro: Haha good point. I was debating between the two names and ended up arbitrarily with `bayesian`. At every step we essentially assign reads that have P(mismatches to a candidate consensus are all sequencing errors | the read actually == the consensus) < some threshold. If you think `probabilistic` fits here better, we can go with that.
One note is that `probabilistic` is sometimes interpreted as "random" (i.e. not deterministic), but this algorithm always returns the same results given the same input.

cassiopeia/preprocess/UMI_utils.py
Lioscro: Yes, this is what is done internally

cassiopeia/preprocess/UMI_utils.py
Lioscro: I wonder how this got through `black`... Perhaps i forgot to run it

cassiopeia/preprocess/UMI_utils.py
Lioscro: I don't expect that to ever happen intentionally. It probably doesn't make much sense to have proportion > 0.5, though I'm not sure what the exact cutoff should be.

cassiopeia/preprocess/UMI_utils.py
Lioscro: We do have a `PreprocessError` defined in `preprocess.py`, but we can't import this into `UMI_utils` because it'll cause a circular import. Do you think its fine to re-define a new error with the same name in `UMI_utils.py`?

cassiopeia/preprocess/UMI_utils.py
Lioscro: You can think of the `q_threshold` and `proportion` as the following.
Assign a read to a candidate consensus if
```
(sum of the quality values where the read does not match the consensus) <= (q_threshold * proportion * length)
```
So, you are conceptually saying I want to allow at most (proportion * length) mismatches, each of base quality q_threshold for all reads assigned to a consensus.

cassiopeia/preprocess/UMI_utils.py
mattjones315: I think to make frame this as Bayesian, we have to be clear what the hypothesis, model, and likelihoods are. In my eyes, the null hypothesis is that the read is _not_ the consensus sequence and the alternative hypothesis is that the read is the consensus sequence. The data we're using is the number and quality of mismatches. So, the probability P(mismatches | read is the consensus) sounds a lot more like a likelihood [the posterior would be P(read is the consensus | mismatches) here]. Does this make sense? And given this, would it make more sense to maybe call this procedure `form_cluster_likelihood`?

cassiopeia/preprocess/UMI_utils.py
mattjones315: Good to know! Maybe we can make this a bit more clear in the documentation then?

cassiopeia/preprocess/UMI_utils.py
mattjones315: Sometimes `black` doesn't handle these things well haha. I sometimes put math on a separate line.

Lioscro: Huh, seems black doesn't want to line break here. will do manually

cassiopeia/preprocess/UMI_utils.py
mattjones315: Sounds good. Should we throw a warning then if the proportion exceeds 0.5?

cassiopeia/preprocess/UMI_utils.py
mattjones315: What if we make a new file called `_errors_and_warnings.py` that we can import from? If you think this is stupid, we can just use the `UMIUtilsError` that you defined here.

cassiopeia/preprocess/UMI_utils.py
Lioscro: Yes! I like how that sounds better than `probabilistic`

cassiopeia/preprocess/UMI_utils.py
Lioscro: Yep, I removed the max,min part because that is just an implementation detail to deal with cases where only one sequence is provided. Also added description to what `q_threshold` is (PHRED quality score).

cassiopeia/preprocess/UMI_utils.py
Lioscro: That's a good idea

cassiopeia/preprocess/UMI_utils.py
```diff
@@ -192,43 +197,69 @@ def form_collapsed_clusters(
...
         UMI_key: A function that takes an alignment and returns the UMI sequence
+        method: Which method to use to form initial sequence clusters. Must be
+            one of the following:
+            * cutoff: Uses a quality score hard cutoff of 30, and any mismatches
+                below this quality are ignored. Initial sequence clusters are
+                formed by selecting the most common base at each position (with
+                quality at least 30).
+            * bayesian: Utilizes the error probability encoded in the quality
+                score. Initial sequence clusters are formed by selecting the
+                most probable at each position.
```
mattjones315, line 216: Need to rename this to `likelihood`

cassiopeia/preprocess/pipeline.py
```diff
@@ -190,6 +188,15 @@ def collapse_umis(
...
             allowed between the sequences of 2 aligned segments to be collapsed
+        method: Which method to use to form initial sequence clusters. Must be
+            one of the following:
+            * cutoff: Uses a quality score hard cutoff of 30, and any mismatches
+                below this quality are ignored. Initial sequence clusters are
+                formed by selecting the most common base at each position (with
+                quality at least 30).
+            * bayesian: Utilizes the error probability encoded in the quality
+                score. Initial sequence clusters are formed by selecting the
+                most probable at each position.
```
mattjones315, line 199: need to change this to `likelihood`

mattjones315 (changes requested):
This looks awesome! Great job addressing these things so quickly. Just one little documentation issue that we need to fix before merging in.

mattjones315 (approved):
LGTM!

PR #106, opened 2021-07-13
preprocessing pipeline improvements

Lioscro:
Depends on #104 and #105. Note that this branch has #105 as its upstream.

This PR implements several improvements to the preprocessing pipeline.
* Full support for ambiguous (conflicting) alleles. A new pipeline option `allow_allele_conflicts` has been added, which allows allele conflicts to be retained (not filtered out). Specifically, this option affects the behaviors of `pp.error_correct_umis` and `pp.filter_molecule_table`. Functions in `pp.utilities` have been appropriately modified to support allele conflicts.
* intBCs can now be corrected to a whitelist, using Levenshtein distance. Note that Levenshtein distance is calculated using a custom implementation in the `ngs_tools` library, which has knowledge of ambiguous bases, unlike the `Levenshtein` library, which does not.
* Implemented a new step in the preprocessing pipeline: `filter_bam`, which filters any reads with low quality bases in either the UMI or barcode sequences.
* Deprecated `skip_existing` option from `pp.collapse_umis` as this behavior is inconsistent with other preprocessing functions. We may want to add this functionality back later, for all preprocessing functions.
* Parallelized the following preprocessing stages: `collapse_umis`, `align_sequences`, `error_correct_umis`
* Moved `n_threads` as a general setting in the preprocessing pipeline.
* Minor change to `data.utilities.compute_dissimilarity_map` so that the internal function can still be numbaized in object mode for a slight speedup.
* Moved `is_ambiguous_state` function from `data.utilities` to `mixins.utilities` because this function is needed in the preprocessing pipeline as well.
* Raise error when trying to use a `GreedySolver` or `ILPSolver` on a tree with ambiguous states.
* The table index (first column containing row index) is no longer written to the output tables in the preprocessing pipeline.


cassiopeia/preprocess/pipeline.py
```diff
@@ -577,22 +624,81 @@ def call_alleles(
     return alignments
 
 
+def error_correct_intbcs(
```
mattjones315, line 627: I would prefer to call this something else because we already have an `error_correct_intbc` step that can be done on molecule tables, correcting intBCs to one another if they share the same allele and cellBC. (This is in `cassiopeia/preprocess/utilities.py` and invoked in `pipeline.filter_molecule_table`)

Why don't we call this `error_correct_intbc_to_whitelist`? 

The naming convention in the setup script and configuration files should also be amended to reflect whatever name you choose.

mattjones315 (changes requested):
This is looking great!

I've left some comments that I'd like to get addressed before moving on. One major (non-technical) issue is on the naming of the `error_correct_intbc` function for whitelist error correction. We already have intBC error correction functionality that takes place in the `filter_molecule_table` step, so we have to disambiguate these. Let's make sure these changes are reflected in the pipeline setup & config files too.

Also, my test suite is complaining because there are two files named `utilities_test.py` - one in `mixin_tests` and one in `data_tests`. Can you rename these to `mixin_utilities_test.py` and `data_utilities_test.py`, respectively?

cassiopeia/preprocess/pipeline.py
```diff
@@ -321,7 +363,7 @@ def resolve_umi_sequence(
     second_reads = {}
     first_reads = {}
 
-    unique_pairs = molecule_table.groupby(["cellBC", "UMI"])
+    unique_pairs = molecule_table.groupby(["cellBC", "UMI"], sort=False)
```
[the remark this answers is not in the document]
Lioscro: Yep, its faster :)

richardyz98 (changes requested):
Hey Joseph,
This looks great to me! Very clean and concise, and great additions.

cassiopeia/preprocess/cassiopeia_preprocess.py
```diff
@@ -70,36 +73,43 @@ def main():
                 f"stage `{entry_point}`"
             )
 
-        if entry_point in ("error_correct_barcodes", "collapse"):
+        if entry_point in ("filter", "error_correct_barcodes", "collapse"):
```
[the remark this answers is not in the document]
Lioscro: Haha good catch! I must have missed it while renaming some stages.

richardyz98 (approved):

cassiopeia/preprocess/__init__.py
```diff
@@ -6,6 +6,8 @@
     align_sequences,
     call_alleles,
     collapse_umis,
+    error_correct_barcodes_to_whitelist,
```
mattjones315, line 9: Are we worried that users will be confused by what `barcodes` means here, thinking it might mean `intBC` or `cellBC`?

cassiopeia/preprocess/pipeline.py
```diff
@@ -97,20 +98,73 @@ def convert_fastqs_to_unmapped_bam(
...
+        n_threads=n_threads,
+    )
+    return filtered_fp
+
+
+def error_correct_barcodes_to_whitelist(
+    bam_fp: str,
+    output_directory: str,
+    whitelist_fp: Optional[str] = None,
+    whitelist: Optional[List[str]] = None,
```
mattjones315, line 151: Let's group these two parameters together into a single one `whitelist` that can either be a file path or a list of intBC strings. 

Also, if there's not a test yet that tests this list-intBC functionality let's add one.

cassiopeia/preprocess/pipeline.py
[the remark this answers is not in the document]
mattjones315: Nice!

cassiopeia/preprocess/pipeline.py
```diff
@@ -577,22 +645,99 @@ def call_alleles(
     return alignments
 
 
+def error_correct_intbcs_to_whitelist(
+    input_df: pd.DataFrame,
+    whitelist_fp: Optional[str] = None,
+    whitelist: Optional[List[str]] = None,
```
mattjones315, line 651: Same as above - let's group these two parameters together into a single `whitelist` parameter, and add a test to make sure this is supported.

mattjones315 (changes requested):
Looking good! I have one or two comments that I'd like to see addressed here before merging it in.

cassiopeia/preprocess/__init__.py
Lioscro: we can rename to `error_correct_cellbcs_to_whitelist`

mattjones315 (approved):
Looks great! Let's merge it in.

PR #111, opened 2021-07-23
Collapsing

richardyz98:
Hi Matt,
Here is the PR standardizing the collapsing of mutationless edges across all algorithms.

The tests are changed to reflect this, and now each algorithm has cases with and without collapsing mutationless edges

Additionally, I've added collapsing of unifurcations to ilp and hybrid solvers, as well as 2 small bugfixes:
1) Robinson Foulds in the critique library now critiques a copy of the tree passed
2) Hybrid now updates self.__tree so that the solver instance is reusable

I've also blacked some old files.

Thanks!


mattjones315:
Thanks Richard! Before I will review this request, I'd like all the tests to be passing. Unfortunately, I get an error when I try to run tests, occurring in `test/mixin_tests/mixing_utilities_test.py`. It appears that `from cassiopeia.mixins import utilities` does not work. Can you fix this so I can review the PR?

cassiopeia/solver/HybridSolver.py
```diff
@@ -171,16 +178,18 @@ def solve(
...
-            cassiopeia_tree.get_tree_topology(), character_matrix
-        )
-        cassiopeia_tree.populate_tree(samples_tree)
+        # collapse mutationless edges
+        if collapse_mutationless_edges:
+            cassiopeia_tree.collapse_mutationless_edges(
+                infer_ancestral_characters=True
+            )
+
+        self.__tree = nx.DiGraph()
```
mattjones315: I know I wrote this, but I wonder if there is a more elegant way to do to this than have a hidden variable `__tree` in the object.

I have a solution that works by creating a temporary tree in the `solve` method & passing it to the top-solver; I will push it to this PR.

test/solver_tests/ilp_solver_test.py
```diff
@@ -313,16 +313,34 @@ def test_ilp_solver_perfect_phylogeny(self):
...
+                            str(
+                                tuple(
+                                    self.duplicates_tree.character_matrix.loc[
+                                        e[1]
+                                    ]
+                                )
+                            ),
+                        )
+                    )
+                else:
```
mattjones315, line 340: I don't think we need all of this logic. If this is because we have hanging leaves on the ILPSolver that aren't part of the sample input, then we actually have a small bug on our hands and should amend the solver post-processing code.

test/solver_tests/ilp_solver_test.py
```diff
@@ -421,16 +439,34 @@ def test_ilp_solver_with_duplicates(self):
...
+                            e[0],
+                            str(
+                                tuple(
+                                    self.duplicates_tree.character_matrix.loc[
+                                        e[1]
+                                    ]
+                                )
+                            ),
+                        )
+                    )
```
mattjones315, line 465: same as above.

mattjones315 (changes requested):
Thanks for the great PR here @richardyz98! I've made some comments that I'd like to address before merging in.

One general concern I have that I'll raise here rather than in specific comments is that we have a general lack of tests making sure that `collapse_mutationless_edges=False` works as expected. Since this PR is about collapsing let's make sure we're checking that both settings work as we'd like. 

Thank you again!

test/solver_tests/ilp_solver_test.py
richardyz98: hi Matt,
This was added because after collapsing unifurcations on the final ILP reconstructed tree we need a new calculation of the parsimony, as we need to add the parsimony from the leaves which are now not attached to an internal node representing their character vector.

test/solver_tests/ilp_solver_test.py
richardyz98: same as above

cassiopeia/solver/HybridSolver.py
richardyz98: Thanks Matt!

richardyz98:
Hi Matt,
I did try to add test cases for each solver which include collapsing and not collapsing edges. These edit the existing tests, so I understand if it's hard to find. I can add additional comments/documentation, or let me know if you feel these tests aren't sufficient.

test/solver_tests/ilp_solver_test.py
mattjones315: I've fixed this in a new push to the PR.

mattjones315 (approved):
Looking great! I'm going to merge it in.

mattjones315 (approved):

PR #119, opened 2021-07-30
Ilp hybrid fixes

richardyz98:
Hi all,
This PR contains fixes to the ILP that renames the internal files such that we can convert trees reconstructed with the ILPSolver into newicks. 

I also added some additional testing for the ILP and Hybird Solvers around logging and newick conversion


cassiopeia/solver/HybridSolver.py
```diff
@@ -9,7 +9,9 @@
 of a VanillaGreedySolver stacked on top of a ILPSolver instance.
 """
 from typing import Dict, List, Generator, Optional, Tuple
+import warnings
```
mattjones315: these aren't being used anymore, I'm removing them

test/solver_tests/hybrid_solver_test.py
```diff
@@ -39,6 +38,16 @@ def find_triplet_structure(triplet, T):
...
+    def assertIsFile(self, path):
+        if not pl.Path(path).resolve().is_file():
+            raise AssertionError("File does not exist: %s" % str(path))
+
+    def assertIsNotFile(self, path):
+        if pl.Path(path).resolve().is_file():
+            raise AssertionError(
+                "File exists when it should not: %s" % str(path)
+            )
+
```
mattjones315, line 50: Are we using this function at all? If not, let's get rid of it.

mattjones315 (approved):
Looks good! Just left one comment that should be resolved before merging in. Else, feel free to merge it in once the build passes.

PR #124, opened 2021-08-03
Branch length estimation: MLE

sprillo:
This PR implements a simple branch length estimation model: the MLE under a model of IID CRISPR/Cas9 memoryless cutting. (See the class docstring for further details.)

I added a few methods to the CassiopeiaTree that are leveraged by the solver.

Test plan: All new code is tested, coverage hits 100%.


cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1872,6 +1872,105 @@ def get_distances(
...
+            CassiopeiaTreeError if the tree has not been initialized.
+        """
+        self.__check_network_initialized()
+
+        current_times = self.get_times().values()
+        max_time = max(current_times)
+        min_time = min(current_times)
+        new_times = {}
+        for node in self.nodes:
+            new_times[node] = self.get_time(node) / (max_time - min_time)
```
mattjones315, line 1892: Isn't this supposed to be `(time - min) / (max - min)`? Or is this functionally the same?

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1872,6 +1872,105 @@ def get_distances(
...
+
+        The longest path from root to leaf will have length 1 after the
+        scaling.
+
+        Raises:
+            CassiopeiaTreeError if the tree has not been initialized.
+        """
+        self.__check_network_initialized()
+
+        current_times = self.get_times().values()
```
mattjones315, line 1887: Is there ever a case when the values are not initialized?

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1872,6 +1872,105 @@ def get_distances(
...
+        max_time = max(current_times)
+        min_time = min(current_times)
+        new_times = {}
+        for node in self.nodes:
+            new_times[node] = self.get_time(node) / (max_time - min_time)
+        self.set_times(new_times)
+
+    def get_depth(self) -> float:
```
mattjones315, line 1895: How is this function different from `get_max_depth_of_tree`?

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1872,6 +1872,105 @@ def get_distances(
...
+        parent_states = self.get_character_states(parent)
+        child_states = self.get_character_states(child)
+        res = []
+        for i, (parent_state, child_state) in enumerate(
+            zip(parent_states, child_states)
+        ):
+            if (
+                parent_state == 0
+                and child_state != 0
+                and child_state != self.missing_state_indicator
```
mattjones315, line 1942: Why not be a bit more general here?

I think the following check would have the same effect and yet be general to reversible models:
```
if parent_state != child_state
and child_state != self.missing_state_indicator
and parent_state != self.missing_state_indicator:
```

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1872,6 +1872,105 @@ def get_distances(
...
+        return res
+
+    def get_unmutated_characters_along_edge(
+        self, parent: str, child: str
+    ) -> List[int]:
+        """
+        List of unmutated characters along edge.
+
+        A character is considered to not mutate if it goes from the zero state
+        to the zero state.
```
mattjones315, line 1954: Once again -- I wonder if we can be a bit more general.

I understand that we can't be as general as we'd like (returning all the characters that don't change along an edge) but what if we introduce an argument that _allows_ us to be that general? This would be along the lines of saying whether or not the mutation model is reversible or not.

cassiopeia/mixins/errors.py
```diff
@@ -77,3 +77,12 @@ class UnknownCigarStringError(Exception):
...
 class UnspecifiedConfigParameterError(Exception):
     pass
+
+
+class BranchLengthEstimatorError(Exception):
+    """An Exception class for the BranchLengthEstimator class."""
+    pass
+
+
+class IIDExponentialMLEError(BranchLengthEstimatorError):
```
mattjones315, line 87: Why do we need a specific error for the IIDExpenentialMLE class? Is this best practice?

cassiopeia/tools/branch_length_estimator/BranchLengthEstimator.py
```diff
@@ -0,0 +1,26 @@
...
+from cassiopeia.data import CassiopeiaTree
+
+
+class BranchLengthEstimator(abc.ABC):
+    """
+    BranchLengthEstimator is an abstract class that all branch length
+    estimation algorithms derive from. At minimum, all BranchLengthEstimator
+    subclasses will implement a method called `estimate_branch_lengths`.
+    """
+
```
mattjones315, line 18: Are there going to be any general variables for the BranchLengthEstimator class such that we want to add this to an abstract `__init__`?

cassiopeia/tools/branch_length_estimator/IIDExponentialMLE.py
```diff
@@ -0,0 +1,174 @@
...
+            since the MLE tends to collapse mutationless edges to length 0.
+        solver: Convex optimization solver to use. Can be "SCS", "ECOS", or
+            "MOSEK". Note that "MOSEK" solver should be installed separately.
+        verbose: Verbosity level.
+
+    Attributes:
+        log_likelihood: The log-likelihood of the training data under the
+            estimated model.
+        mutation_rate: The estimated CRISPR/Cas9 mutation rate, assuming that
+            the tree has depth exactly 1.
```
mattjones315, line 49: Should these attributes be initialized in the `__init__`? I wonder if you want to throw an error if a user tries to access these values before a model has been run or if it should just return `None`?

cassiopeia/tools/branch_length_estimator/IIDExponentialMLE.py
```diff
@@ -0,0 +1,174 @@
...
+        if solver not in allowed_solvers:
+            raise ValueError(
+                f"Solver {solver} not allowed. Allowed solvers: {allowed_solvers}"
+            )  # pragma: no cover
+        self._minimum_branch_length = minimum_branch_length
+        self._verbose = verbose
+        self._solver = solver
+
+    def estimate_branch_lengths(self, tree: CassiopeiaTree) -> None:
+        r"""
```
mattjones315, line 68: Our code best-practices are to start with a single-line (less than 80 characters) docstring header.

cassiopeia/tools/branch_length_estimator/IIDExponentialMLE.py
```diff
@@ -0,0 +1,174 @@
...
+
+        # # # # # Extract log-likelihood # # # # #
+        log_likelihood = float(log_likelihood.value)
+        if np.isnan(log_likelihood):
+            log_likelihood = -np.inf
+        self._log_likelihood = log_likelihood
+
+        # # # # # Extract mutation rate # # # # #
+        tree_depth = tree.get_depth()
+        self._mutation_rate = tree_depth
```
mattjones315, line 153: I misunderstood what this parameter is supposed to represent. Why is it the maximum depth of the tree and not the probability that a site will be cut in a given time step? Are these the same thing here?

cassiopeia/tools/branch_length_estimator/IIDExponentialMLE.py
```diff
@@ -0,0 +1,174 @@
...
+        )
+
+        # # # # # Create constraints of the optimization problem # # # # #
+        a_leaf = tree.leaves[0]
+        root = tree.root
+        root_has_time_0_constraint = [r_X_t_variables[root] == 0]
+        minimum_branch_length_constraints = [
+            r_X_t_variables[child]
+            >= r_X_t_variables[parent]
+            + minimum_branch_length * r_X_t_variables[a_leaf]
```
mattjones315, line 96: I'm not sure I understand why you need to enforce that `time(child) >= time(parent) + min_branch_length + time(a_leaf)`. Specifically, why do you need to have the term `time(a_leaf)` here in this constraint? Thank you in advance for elucidating this!

In general, perhaps it would be good to make clear these constraints of the MLE problem.

test/data_tests/cassiopeia_tree_test.py
```diff
@@ -1664,6 +1664,35 @@ def test_set_dissimilarity(self):
                 "node3", {leaf: 0 for leaf in tree_no_map.leaves}
             )
 
+    def test_get_depth(self):
+        tree = cas.data.CassiopeiaTree(tree=self.test_network)
+        self.assertEqual(tree.get_depth(), 8)
+
```
mattjones315, line 1670: I'm pretty sure this operates exactly as `get_max_depth_of_tree`

test/tools_tests/branch_length_estimator_tests/iid_exponential_mle_test.py
```diff
@@ -0,0 +1,363 @@
...
+        self.assertTrue(0.9 < tree.get_time("5") < 1.1)
+        self.assertTrue(0.9 < tree.get_time("6") < 1.1)
+        self.assertTrue(1.4 < model.mutation_rate < 1.6)
+        self.assertAlmostEqual(tree.get_time("0"), 0.0, places=3)
+
+    @parameterized.expand([("ECOS", "ECOS"), ("SCS", "SCS")])
+    def test_subtree_collapses_when_no_mutations(self, name, solver):
+        """
+        A subtree with no mutations should collapse to 0. It reduces the problem to
+        the same as in 'test_hand_solvable_problem_1'
```
mattjones315, line 291: This test is failing for me:

```
self.assertAlmostEqual(model.log_likelihood, -1.386, places=3)
E   AssertionError: -inf != -1.386 within 3 places
```

mattjones315 (changes requested):
Fantastic PR Sebastian! Thanks so much for the work here.

I've left some comments throughout that I'd like to get resolved before we submit it. Perhaps most salient is the redundancy of a function in the `CassiopeiaTree` class (`get_depth`, I believe is not needed) and a failing test in `iid_exponential_mle_test.py`.

I look forward to reviewing the revisions and merging this in soon!

test/tools_tests/branch_length_estimator_tests/iid_exponential_mle_test.py
mattjones315: I should be more specific - it looks like it works with ECOS but fails with SCS.

cassiopeia/data/CassiopeiaTree.py
sprillo: Either way, the tree will have unit length after the scaling. However, I like your proposal better because it makes the root have time 0, so I will push this change.

Note that if we were to enforce that throughout the lifetime of the tree the root has time zero, this code can be simplified to `time  / max`. But I don't think we are enforcing this constraint, not it's clear to me that we should.

cassiopeia/data/CassiopeiaTree.py
sprillo: Only if the network is not initialized, which is checked for in the previous line, or am I missing something?

cassiopeia/data/CassiopeiaTree.py
sprillo: Ah this is really interesting:

1) I hadn't noticed that the method `get_max_depth_of_tree` already existed.
2) If the root has time different from 0, `get_max_depth_of_tree` and `get_depth` have different behavior. I believe this is a bug in `get_max_depth_of_tree`.

My proposal is that we remove `get_depth` and fix `get_max_depth_of_tree`.

cassiopeia/data/CassiopeiaTree.py
sprillo: Yeah I like this!

By the way, note that `CassiopeiaTree` already implements a method `get_mutations_along_edge` which assumes that missing characters are mutated characters (so it doesn't have the behavior I need). The returned object is also a `List[Tuple[int, int]]` instead of a `List[int]`.

I don't think that `get_mutations_along_edge` is used anywhere in the codebase. We can try to unify both methods, of keep both, or even remove `get_mutations_along_edge`. What do you think?

cassiopeia/data/CassiopeiaTree.py
sprillo: We could do this, but I don't think we need that behavior now (or ever). In fact I'm realizing that maybe we should also keep `get_mutated_characters_along_edge` as-is for the same reason. I'm just following [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) here.

cassiopeia/tools/branch_length_estimator/IIDExponentialMLE.py
sprillo: The property `mutation_rate` is supposed to represent the _estimated mutation rate of CRISPR/Cas9 under the assumption that the tree has depth 1_ (I will add this to the property's docstring, as it's missing). This is an implementation detail, but let me clarify why it is set to the depth of the tree at this line: Note that internally, what is solved for by the convex optimization solver is not the time of each node in the tree, but rather the time _multiplied_ by the mutation rate. As a consequence, after running the third-party solver, what we have is the time of each node _multiplied_ by the mutation rate. In particular, if we now assume that the tree has depth one, it means that the estimate for the time of a leaf (i.e. the depth of the tree) is equal to the mutation rate. I will re-structure the logic of the code to make this clearer, as I can see that now it is confusing.

cassiopeia/tools/branch_length_estimator/IIDExponentialMLE.py
sprillo: This is because the variables of the convex optimization problem are the times _multiplied_ by the mutation rate. Thus the following are equivalent, under the assumption that the tree has depth equal to 1:

1) `tree.get_time(child) >= tree.get_time(parent) + minimum_branch_length`

2) `r_X_t_variables[child] >= r_X_t_variables[parent] + minimum_branch_length * r_X_t_variables[a_leaf]`

The proof is: just divide by the mutation rate in (2), and use the fact that under the unit-depth assumption we have`r_X_t_variables[a_leaf] = mutation_rate`. You obtain exactly (1).

You might wonder why we need to parameterize the problem in terms of the time _multiplied_ by the rate, instead of just parameterizing the problem in terms of the times and the mutation rate parameter. This is because:
1) The times of the nodes and the mutation rate are not identifiable: they are only identifiable up to their product, so we don't gain anything with this richer parameterization (it has exactly 1 more parameter).
2) The optimization problem in terms of the times and the mutation rate parameter is (perhaps surprisingly) NOT a convex problem, whereas in terms of the times _multiplied_ by the mutation rate it is.

Note that all this is an implementation detail. What is promised by the class' docstring (that the MLE is computed assuming the tree has depth 1 and the branches have length at least `minimum_branch_length`) is exactly what is being done.

(Okay, technically there is at least one exception: when the character matrix is completely unmutated, in which case the MLE exists but I throw an error. This has to do with the fact what when the character matrix is totally unmutated, the MLE of the mutation rate is 0, and so the proof I gave above fails since we are dividing by zero! So: there actually are border cases where I don't do what is promised, but they are so extreme that nobody will ever encounter them in practice.)

test/data_tests/cassiopeia_tree_test.py
sprillo: Addressed in a previous comment.

test/tools_tests/branch_length_estimator_tests/iid_exponential_mle_test.py
sprillo: Yes, this is really weird. The test passes on the cluster (and on Github), but fails on my machine and on yours. I'll look into it.

mattjones315: Could this be a difference in `cvxpy` versions? My version that is failing with is 1.1.7

cassiopeia/data/CassiopeiaTree.py
mattjones315: Good call -- I understand now why you were initially just dividing `time / (max - min)`, but yeah I think we should be more specific here and move to `(time - min) / (max-min)`.

cassiopeia/data/CassiopeiaTree.py
mattjones315: yeah you're right -- if the network is initialized through `populate_tree`, then this should be just fine.

cassiopeia/data/CassiopeiaTree.py
mattjones315: Yes I agree.

Just so we're on the same page -- the bug you're referring to is the fact that `get_max_depth_of_tree` need not return the maximum distance _from the root_ but just the maximum time of any leaf?

Then, the fix should be pretty easy -- we just want to replace the list comprehension statement with:

```
root_time = self.get_time(self.root)
depths = [self.get_time(l) - root_time for l in self.leaves]
```

cassiopeia/data/CassiopeiaTree.py
mattjones315: Good catch! I had forgotten that we had implemented this function. As it is, I think you're right - we're not using `get_mutations_along_edge` explicitly in the codebase but from my previous work analyzing these trees, I actually prefer returning `List[Tuple[int, int]]`. Sometimes you just want the mutated characters (like you are) and sometimes you actually want the mutated states (e.g., for computing likelihoods).

So, I would prefer to remove your function `get_mutated_characters_along_edge` and adapt the existing `get_mutations_along_edge` to your specifications by adding extra arguments.

cassiopeia/data/CassiopeiaTree.py
mattjones315: Haha, I hadn't heard of YAGNI before but I love it.

You're right -- let's not extend this function to the general case until we're sure we need it.

cassiopeia/tools/branch_length_estimator/IIDExponentialMLE.py
mattjones315: Ah that makes a lot of sense! For what it's worth, I think we were on the same page as to what the `mutation_rate` variable was supposed to store but I was unsure why you were setting it the way you were. Thanks so much for the clarification here. 

Perhaps making this a bit more clear in the code would be nice for future users struggling to understand that line of code.

cassiopeia/data/CassiopeiaTree.py
sprillo: Yes, I agree!

cassiopeia/tools/branch_length_estimator/IIDExponentialMLE.py
mattjones315: This makes a lot more sense, especially after reading your response to my question about how you set `mutation_rate`. Thanks again for walking me through this!

cassiopeia/mixins/errors.py
sprillo: It is common practice: subclassing exceptions allows one to be more specific when catching exceptions.

cassiopeia/tools/branch_length_estimator/BranchLengthEstimator.py
sprillo: I don't think so, at least not right now. I'd say if any repeated code comes up in the future that is clearly an essential part of a BranchLengthEstimator's init we can add it at that time.

cassiopeia/tools/branch_length_estimator/IIDExponentialMLE.py
sprillo: This is a good point. I think it's best practice to instantiate all relevant attributes in __init__, so I'll do that.

mattjones315:
@sprillo -- I also noticed that you have put your tests under `tools_test` but the recent PR containing the `small_parsimony.py` library has put tests under `tool_test`. Can you reconcile these differences?

cassiopeia/tools/branch_length_estimator/IIDExponentialMLE.py
sprillo: Yes, I will restructure the logic.

cassiopeia/mixins/errors.py
mattjones315: Sounds good!

cassiopeia/tools/branch_length_estimator/BranchLengthEstimator.py
mattjones315: Great, let's proceed as you suggest.

cassiopeia/data/CassiopeiaTree.py
sprillo: Done

sprillo:
@mattjones315 I think it is best to do it in the other PR ( https://github.com/YosefLab/Cassiopeia/pull/125 ), renaming `tool_test` to `tools_test` (since the module is called `tools` after all). Could you do that please since you own that PR?

sprillo:
@mattjones315 All comments should be addressed (I resolved the conversations, should I have let you do that, since you are the reviewer?)

sprillo:
> I also noticed that you have put your tests under tools_test but the recent PR containing the small_parsimony.py library has put tests under tool_test. Can you reconcile these differences?

Oh, I see, it's a different PR that already got merged in. I can rename the directory then.

test/data_tests/cassiopeia_tree_test.py
```diff
@@ -1664,6 +1664,40 @@ def test_set_dissimilarity(self):
                 "node3", {leaf: 0 for leaf in tree_no_map.leaves}
             )
 
+    def test_scale_to_unit_length(self):
```
mattjones315, line 1667: Can you add an additional test to verify that `get_max_depth_of_tree` works as described now?

mattjones315 (changes requested):
Looking great! I just have one more request -- since we've noticed some nuances to `get_max_depth_of_tree`, can you add a test verifying that we're doing what we advertise?

Also, could you change the function `get_mean_depth_of_tree` to match the behavior we have in `get_max_depth_of_tree`?

test/data_tests/cassiopeia_tree_test.py
sprillo: Great point. Now `get_max_depth_of_tree` and `get_mean_depth_of_tree` should be correct and tested for our border case. I modified the already existing `test_depth_calculations_on_tree` test to accomplish this.

mattjones315 (approved):
Looks great! Let's merge this in. Thanks for the great work here Sebastian!

PR #125, opened 2021-08-03
FitchCount

mattjones315:
This PR implements FitchCount, previously described in [Quinn et al, Science 2021](https://science.sciencemag.org/content/early/2021/01/21/science.abc1944). This is implemented as a function in `cassiopeia/tools/small_parsimony.py`, but per a conversation with @sprillo, we are entertaining the idea of eventually moving this over to a separate library with an emphasis on rate-matrix estimation.

As I implemented FitchCount, I also implemented functionality for performing a breadth-first-search on the tree, as well as subsetting the tree to get the subtree below a given node.


cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1116,6 +1116,30 @@ def depth_first_traverse_edges(
 
         return nx.dfs_edges(self.__network, source=source)
 
+    def breadth_first_traverse_edges(
+        self, source: Optional[int] = None
```
sprillo, line 1120: I believe you mean `Optional[str]`, since nodes `str` in `CassiopeiaTree`?

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1144,6 +1168,37 @@ def leaves_in_subtree(self, node) -> List[str]:
...
+        nodes_in_subtree = self.depth_first_traverse_nodes(source=node)
+        to_remove = list(set(self.nodes) - set(nodes_in_subtree))
+        for node in to_remove:
+            self.__remove_node(node)
+
+        # reset cache because we've changed the tree topology
+        self.__cache = {}
+
+        # Remove all removed nodes from data fields
+        self.__register_data_with_tree()
```
sprillo, line 1200: Why don't we make `__register_data_with_tree` clear the cache? Is there ever a time one would `__register_data_with_tree` and not clear the cache? Sounds like a source of bugs. The other times `__register_data_with_tree ` is used is in `add_leaf` and `remove_leaf_and_prune_lineage`, where the same pattern as in this PR is used. This leads to a duplicated pattern of `self.__cache = {}` followed by `self.__register_data_with_tree()`.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1874,4 +1929,4 @@ def get_distances(
 
     def copy(self) -> "CassiopeiaTree":
         """Full copy of CassiopeiaTree"""
-        return copy.deepcopy(self)
\ No newline at end of file
+        return copy.deepcopy(self)
```
sprillo: This is unrelated to this PR, but I think the standard way to implement copy and deepcopy in a class is by implementing the `__copy__` and `__deepcopy__` methods. In other words here we want to define `__deepcopy__`, not `copy`. See https://docs.python.org/3/library/copy.html , specifically:

_In order for a class to define its own copy implementation, it can define special methods \_\_copy\_\_() and \_\_deepcopy\_\_(). The former is called to implement the shallow copy operation; no additional arguments are passed. The latter is called to implement the deep copy operation; it is passed one argument, the memo dictionary. If the \_\_deepcopy\_\_() implementation needs to make a deep copy of a component, it should call the deepcopy() function with the component as first argument and the memo dictionary as second argument._

cassiopeia/tools/small_parsimony.py
```diff
@@ -239,3 +241,235 @@ def score_small_parsimony(
...
+    cassiopeia_tree: CassiopeiaTree,
+    meta_item: str,
+    root: Optional[str] = None,
+    infer_ancestral_states: bool = True,
+    state_key: str = "S1",
+):
+    """Runs the FitchCount algorithm.
+
+    Performs the FitchCount algorithm for inferring the number of times that
+    two two states transition to one another across all equally-parsimonious
```
sprillo, line 256: typo: `two two`

cassiopeia/tools/small_parsimony.py
```diff
@@ -239,3 +241,235 @@ def score_small_parsimony(
...
+            argument specifies what the attribute name is in the CassiopeiaTree
+
+    Returns:
+        An MxM count matrix indicating the number of edges that contained a
+            transition between two states across all equally parsimonious
+            solutions returned by Fitch-Hartigan.
+    """
+    cassiopeia_tree = cassiopeia_tree.copy()
+
+    unique_states = cassiopeia_tree.cell_meta[meta_item].unique()
```
sprillo, line 283: Instead of inferring the states from the table, why not have the user provide them optionally, to cover the case that there is a state with zero instances? This might sound like a strange use case, but for rare states that appear on some trees and not on others I think it's nice for M to be the same for all trees.

cassiopeia/tools/small_parsimony.py
```diff
@@ -239,3 +241,235 @@ def score_small_parsimony(
...
+    To obtain probabilities P(m1 -> m2), divide each row by its row-sum.
+
+    This procedure will only work on categorical data and will otherwise raise
+    an error.
+
+    Args:
+        cassiopeia_tree: CassiopeiaTree object with a tree and cell meta data.
+        meta_item: A column in the CassiopeiaTree cell meta corresponding to a
+            categorical variable.
+        root: Node to treat as the root.
```
sprillo, line 270: Does it mean that the tree is re-rooted at this node, or that only the subtree at this node will be considered?

test/tools_tests/small_parsimony_test.py
```diff
@@ -313,6 +310,68 @@ def test_general_tree_parsimony(self):
...
 
+    def test_fitch_count_basic_binary(self):
+
+        fitch_matrix = cas.tl.fitch_count(self.binary_tree, "nucleotide")
+
+        num_nucleotides = self.binary_tree.cell_meta["nucleotide"].nunique()
+        self.assertEqual(fitch_matrix.shape, (num_nucleotides, num_nucleotides))
+
+        expected_matrix = pd.DataFrame.from_dict(
+            {"A": [9, 2, 1], "G": [0, 2, 0], "C": [0, 0, 0]},
```
sprillo, line 321: Did you compute the solution by hand?

sprillo:
(@mattjones315 just added a couple comments / questions, not really a review!)

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1144,6 +1168,37 @@ def leaves_in_subtree(self, node) -> List[str]:
 
         return self.__cache["subtree"][node]
 
+    def subset_tree_at_node(
```
Lioscro, line 1171: How do you feel about naming this function `clade` or simply `subtree`?

Lioscro (changes requested):
Just one minor comment about name for the subtree function.

cassiopeia/data/CassiopeiaTree.py
mattjones315: thanks!

cassiopeia/data/CassiopeiaTree.py
mattjones315: This is a good point. I'm going to make an issue for this.

cassiopeia/tools/small_parsimony.py
mattjones315: It means that only the subtree below this node is considered. I'll change this in the docs to make it more clear!

cassiopeia/data/CassiopeiaTree.py
mattjones315: I agree -- I've changed the logic there.

cassiopeia/tools/small_parsimony.py
mattjones315: Not a bad idea, I'll add this as an option.

test/tools_tests/small_parsimony_test.py
mattjones315: Yup!

cassiopeia/data/CassiopeiaTree.py
mattjones315: I like it! I'm going to name it `subset_clade`, indicating clearly that we're subsetting the tree to obtain the clade at the node specified by the user.

mattjones315:
Thanks for the help @sprillo & @Lioscro! I believe I've addressed all of your collective comments. @Lioscro -- would you mind taking another look when you have a chance and letting me know if the codebase looks good for merging in or if there's anything else you'd like me to change? Thanks in advance!

Lioscro (approved):
LGTM!

PR #132, opened 2021-08-06
Ilp post processing fix

### LINKED ISSUE #131: Issues with ILP solution post-processing
2 Issues with ILP

1) Some samples are being removed in the tree in post-processing
2) Some spurious (non-sample) leaves remain in the tree, despite our explicit removing of these nodes

richardyz98:
Here is a PR to fix 2 issues in the ILP post-processing and test:

1) We now remove spurious roots first before we prune edges, as previously if we prune edges first we may leave a sample only attached to a spurious root, thus deleting the leaf from the final tree
2) Fixed a bug in the way that we were removing spurious (non-sample) leaves from the final tree after appending samples, as before we were unintentionally not removing some of these leaves due to not handling an edge case when 2 spurious leaves are sisters

Closes #131


cassiopeia/solver/ILPSolver.py
```diff
@@ -674,16 +674,14 @@ def __append_sample_names_and_remove_spurious_leaves(
...
             if l not in character_matrix.index:
-                to_drop.append(l)
-
-                parent = [p for p in solution.predecessors(l)][0]
-                while solution.out_degree(parent) < 2:
-                    to_drop.append(parent)
-                    parent = [p for p in solution.predecessors(parent)][0]
+                curr_parent = list(solution.predecessors(l))[0]
+                solution.remove_node(l)
+                while len(list(solution.successors(curr_parent))) < 1 and curr_parent != root:
```
mattjones315, line 682: I don't think this is blacked.

mattjones315 (changes requested):
Looks like a tight fix and the great test case! Thanks for the work here Richard. All is good to merge in, except I don't think `ILPSolver.py` has been blacked yet.

mattjones315 (approved):
looks good!

Issue #127, opened 2021-08-05
Global Alignment in Cassiopeia Preprocessing


mattjones315:
We've found that the local alignment strategy used in our Cassiopeia preprocessing pipeline is not ideal for some technologies like GESTALT. We'd like to implement a global alignment option in `cas.pp.align_sequences` as was described in the original GESTALT paper ([McKenna et al, 2016](https://science.sciencemag.org/content/353/6298/aaf7907)).

mattjones315:
We've noticed this problem too when there is low sequencing quality at the end of the read as in #139

mattjones315:
This has been fixed in #133 and #142.

As a future note, we should still consider adding a function for removing poly-A tails from sequence reads.

PR #133, opened 2021-08-06
Minor changes to preprocessing utilities

mattjones315:
This PR addresses some issues that we identified in the preprocessing pipeline, especially related to using GESTALT data. These are miscellaneous and include:

* Raising a warning if there is missing data detected in the alleles after `cas.pp.call_alleles`. This likely stems from the alignment strategy not being ideal for GESALT data (see #127).
* Enabling missing data to be specified when converting to character matrices. This allows users to let these functions know how missing data is encoded as an allele and the changes to the utilities will not automatically convert these to `missing_state_indicator`.
* Flexibility to pass in a single lineage group into `cas.pp.lineage_utils.filtered_lineage_group_to_allele_table`.

### DIFF
--- a/cassiopeia/preprocess/lineage_utils.py
+++ b/cassiopeia/preprocess/lineage_utils.py
@@ -388,8 +388,10 @@ def filtered_lineage_group_to_allele_table(
     Returns:
         final_df: A final processed DataFrame with indel information
     """
-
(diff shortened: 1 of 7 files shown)


richardyz98 (approved):
Looks good to me Matt!

PR #135, opened 2021-08-07
Fix clerical issues

richardyz98:
Hi all,
Here we have some additional tests and minor fixes:

1) Testing that errors are thrown when initializing the tree with both the character matrix and the tree. Specifically, we test to make sure that if the leaves of the tree and the character matrix do not agree, an error is thrown.
2) We now throw errors when character states are not annotated, in the getter method (if "character_states" is not an attribute in the underlying networkx object) and in collapsing_mutationless_edges (if nodes have empty lists as their character states, we error as opposed to collapsing the entire tree)
3) Changed the behavior of `UniformLeafSubsampler` to instead of specifying a source of collapse (which may be confusing), to indicate whether or not to collapse a singular edge leading from the root in the returned induced subtree, if it exists. I wanted to get your opinions on whether you think this is the best way to handle this. One alternative is to simply always preserve this edge if it exists, and tell the user that if they want to get rid of it to just collapse unifurcations again on the returned induced subtree. Please let me know your thoughts!

Thanks,
Richard


cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1020,10 +1020,16 @@ def get_character_states(
...
+            CassiopeiaTreeError if the tree has not been initialized or if
+            the character states are not annotated at the specified node.
         """
         self.__check_network_initialized()
 
+        if "character_states" not in self.__network.nodes[node]:
+            raise CassiopeiaTreeError(
+                f"Character states not annotated at node {node}."
+            )
+
```
mattjones315, line 1032: When would this possibly happen?

Any time you attempt to add a tree to a `CassiopeiaTree` object, you'll have to run `populate_tree`, which will by default add character sates to the attributes of each node. If no character states are specified, then an empty list will be used.

I'm not sure that this will ever be hit - can you specify a case where this might happen?

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1607,12 +1613,19 @@ def collapse_mutationless_edges(
...
         """
         if infer_ancestral_characters:
             self.reconstruct_ancestral_characters()
 
         for n in list(self.depth_first_traverse_nodes(postorder=True)):
+            if len(self.get_character_states(n)) == 0:
+                raise CassiopeiaTreeError(
+                    f"Character states empty at node {n}, annotate"
+                    " character states or infer ancestral characters before"
+                    " mutationless edges can be collaped."
```
mattjones315, line 1627: Two things:

(1) First, let's be more clear that you can set `infer_ancestral_states=True` in this method to take care of this issue.
(2) Second, I'm still not sure why the `character_states` attribute would ever be empty at a node since `populate_tree` always gets run.

cassiopeia/simulator/UniformLeafSubsampler.py
```diff
@@ -62,14 +60,24 @@ def subsample_leaves(
         character states, meta data, and the dissimilarity map for the sampled
         cells only.
 
+        Has the option to keep the single edge leading from root in the
```
mattjones315, line 63: typo: `from root` -> `from _the_ root`

cassiopeia/simulator/UniformLeafSubsampler.py
```diff
@@ -62,14 +60,24 @@ def subsample_leaves(
...
+        induced subtree, if it exists. This edge is often used to represent the
+        time that the root lives before any divisions occur in the phyologeny,
+        and is useful in instances where the branch lengths are critical, like
+        simulating ground truth phylogenies or estimating branch lengths.
+
         Args:
             tree: The CassiopeiaTree for which to subsample leaves
-            collapse_source: The source node from which to collapse
-                unifurcations
+            keep_root_unary_edge: Whether or not to collapse the single edge
```
mattjones315, line 71: argument name does not match the one in the function call.

cassiopeia/simulator/UniformLeafSubsampler.py
```diff
@@ -62,14 +60,24 @@ def subsample_leaves(
...
+                leading from the root in the subsample, if it exists
 
         Returns:
             A new CassiopeiaTree that is the induced subtree on a sample of the
-            leaves in the given tree.
+            leaves in the given tree
+
+        Raises:
+            LeafSubsamplerError if the sample size is <= 0, or larger than the
+            number of leaves in the tree
```
mattjones315, line 80: you'll have to indent the second line of the `Returns` section and the `Raises` section.

mattjones315 (changes requested):
Thanks for the PR @richardyz98 ! 

I left a few comments - some aspects of this PR I do not understand. If there was a bug that you were hitting it would be helpful to know about this. As it stands, some of these changes I feel are unnecessary.

Additionally, I noticed that you changed the API call in the cassiopeia tree test suite from `cas.data.CassiopeiaTree`. Why is this? I prefer to keep it this way because if a user wants to get an idea of usage, they can look at this test suite to understand how we intend them to work with Cassiopeia. It sounds like it created a lot of busy work for you to change these imports, so I'm curious if you ran into a specific problem that made you want to do this.

cassiopeia/data/CassiopeiaTree.py
richardyz98: This was a bug indeed. If you specify the tree upon initialization but not the character matrix, some nodes would not have the "character_states" attribute initialized (see the first case in `test_states_no_annotation`)  but now I realize with the help of this comment that it's caused be the fact that we're only annotating the attribute as an empty list for internal nodes and not leaf nodes. We should have it do all nodes and it would fix this
image

cassiopeia/data/CassiopeiaTree.py
richardyz98: So the issue here is that if the user erroneously believes that the internal nodes are annotated when they are indeed not, then if they run `collapse_mutationless_edges` then they'll end up collapsing the entire tree which would definitely mess up their pipelines if they aren't aware of it, so we just spit an error to prevent headache and debugging on their part. What's meant by "character states are empty" here is that the character states are the empty list, not necessarily that the attribute doesn't exist. I'll make it more explicit that they can set the argument "infer_ancestral_states=True" to fix this. Thanks!

richardyz98: This case is tested in the second case of `test_states_no_annotation`

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -178,9 +178,12 @@ def populate_tree(
...
         if character_matrix is not None:
             self.set_character_states_at_leaves(layer=layer)
+            for n in self.internal_nodes:
+                self.__network.nodes[n]["character_states"] = []
 
-        for n in self.internal_nodes:
-            self.__network.nodes[n]["character_states"] = []
+        else:
+            for n in self.nodes:
+                self.__network.nodes[n]["character_states"] = []
```
mattjones315, line 186: It'd be more parsimonious to do:

```
for n in self.nodes:
    self.__network.nodes[n]['character_states'] = []

if character_matrix is not None:
    self.set_character_states_at_leaves(layer=layer)
```

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -962,7 +963,7 @@ def set_character_states(
         if self.is_leaf(node):
             if self.get_character_states(node) == []:
                 raise CassiopeiaTreeError(
-                    "Leaf node character states have not been instantiated"
+                    "Character states have not been initialized at leaves."
```
mattjones315, line 966: Can you tell people how to initialize character states at leaves in this error message?

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1607,14 +1608,27 @@ def collapse_mutationless_edges(
                 of the tree
 
         Raises:
-            CassiopeiaTreeError if the tree has not been initialized.
+            CassiopeiaTreeError if the tree has not been initialized or if
+            a node does not have character states
```
mattjones315, line 1612: Need to indent second line to make docs look alright.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1607,14 +1608,27 @@ def collapse_mutationless_edges(
...
         """
         if infer_ancestral_characters:
             self.reconstruct_ancestral_characters()
 
         for n in list(self.depth_first_traverse_nodes(postorder=True)):
             if self.is_leaf(n):
+                if self.get_character_states(n) == []:
+                    raise CassiopeiaTreeError(
+                        "Character states have not been initialized at leaves."
```
mattjones315, line 1621: ditto - tell people how to do this if they hit this error message.

mattjones315 (changes requested):
Thanks Richard! I left you a few comments that I'd like to get addressed.

mattjones315 (approved):
Looks good!

mattjones315:
@richardyz98 -- can you please merge this into master and close this PR? Remember to first merge in the new changes from master before merging this in.

Thanks!

PR #137, opened 2021-08-10
Cleanup filtering code and behavior in preprocess

Lioscro:
I noticed there were a few inconsistencies between the variable name & docstring vs the actual filtering behavior in `pp.utilities.filter_cells` and `pp.utilities.filter_umis`. This PR addresses these inconsistencies.

* Renamed the `umi_read_thresh` argument  to `pp.filter_molecule_table` to `min_reads_per_umi` to match the naming of the other variables. The corresponding argument to `pp.utilities.filter_umis` has also been changed.
* All filter thresholds (`min_umi_per_cell`, `min_avg_reads_per_umi`, `min_reads_per_umi`) specifies the minimum number required for a cell/UMI to *pass* filtering. Previously, when a condition equaled any one of these values, the cell/UMI would be filtered out (due to the use of a `<=` operator.
* Removed the second (redundant) filtering step that was not actually performing any additional filtering in `pp.filter_molecule_table`.
* Updated filtering code in `pp.utilities.filter_cells` to perform pandas operations more efficiently by using `groupby` operations.

Overall, these changes result in around 10x speedup of the `pp.filter_molecule_table` function.

### DIFF
--- a/cassiopeia/preprocess/UMI_utils.py
+++ b/cassiopeia/preprocess/UMI_utils.py
@@ -230,11 +230,13 @@ def form_collapsed_clusters(
         max_read_length = max(max_read_length, al.query_length)
 
     # Raise warning when max_hq_mismatches / max_read_length > 0.5
-    warnings.warn(
-        "Provided `max_hq_mismatches` exceeds half of the maximum read length. "
-        "Most reads will be collapsed into a single consensus sequence.",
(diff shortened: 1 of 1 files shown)


mattjones315 (approved):
Thanks for doing this Joseph! This all looks good to me. Let's go ahead and merge it in.

Issue #118, opened 2021-07-29
reconstruct-lineage error: IndexError: index 2 is out of bounds for axis 0 with size 2


isshamie:
Hello I'm trying to run reconstruct_lineages.ipynb but with the test data data/test_at.txt as the allele table. It seemed to output the allele table fine  but when I ran reconstruct lineage I got an error. I am wondering what is the exact input format for the reconstruct-lineage section? Right now the test_lg4_character_matrix.txt I have is below (just first 3 lines). Below that is the error I encountered. Thank you.
```
cellBC	r0	r1	r2	r3	r4	r5	r6	r7	r8	r9	r10	r11	r12	r13	r14	r15	r16	r17	r18	r19	r20	r21r22	r23	r24	r25	r26	r27	r28	r29	r30	r31	r32	r33	r34	r35	r36	r37	r38	r39	r40	r41	r42	r43	r44	r45	r46	r47	r48	r49	r50

IVLT-2B_00.AAACCTGGTCTGGTCG-1	0	0	0	2	2	0	2	0	0	0	2	2	0	2	0	0	0	0	2	20	0	2	2	0	2	0	2	0	0	0	2	0	0	0	0	2	0	0	-	-	-	-	--	-	-	-	-	-	-  
IVLT-2B_00.AAACCTGTCCACTCCA-1	0	0	0	2	0	0	2	0	2	-	-	-	0	2	0	2	0	0	2	20	0	2	2	2	0	0	2	0	2	-	-	-	2	2	0	0	2	2	0	0	0	-	--	-	-	-	-	-	-  
```

The command and error message
```
reconstruct-lineage test_lg4_character_matrix.txt test_lg4_tree.pkl --hybrid
running algorithm...
Using 1 threads, 48 available.
Sending off Target Sets: 1
Started new thread for: 0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0 (num targets = 10) , pid = ea70d39f9eb88a0a5b0739128f47b4d5
(2, 51)
concurrent.futures.process._RemoteTraceback:
"""
Traceback (most recent call last):
  File "/home/isshamie/software/Cassiopeia/cassiopeia/TreeSolver/lineage_solver/lineage_solver.py", line 421, in wrapped
    return func(*args, **kwargs)
[...]

PR #140, opened 2021-08-13
ILPSolverError when Potential Graph cannot be inferred

mattjones315:
We want to throw an informative error when a potential graph cannot be inferred (e.g., if the maximum potential graph layer size is too small for the dataset). This PR address the issue raised recently noticing an uninformative KeyError when the potential graph cannot be inferred (#118).


richardyz98 (approved):
Looks great!

PR #141, opened 2021-08-13
Bulk remove leaf

richardyz98:
Hi,
I noticed that my simulation pipelines were significantly slower, specifically in sub-sampling. I believe this is due to the fact that beforehand, were were calling `register_data_with_tree` between removing each individual node. So now, I've made removing leaves and pruning lineages into a bulk function that handles many leaves and registers the data once at the end.


cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1503,38 +1503,43 @@ def add_leaf(
...
-        the node from all leaf data.
+        Removes the specified leaves and all ancestors of those leaves that are
+        no longer the ancestor of any of the remaining leaves. In the context
+        of a phylogeny, this prunes the lineage of all nodes no longer relevant
+        to observed samples. Additionally, maintains consistency with the
+        updated tree by removing the node from all leaf data.
 
         Args:
-            node: The leaf node to be removed
+            nodes: The leaf (leaves) to be removed
```
mattjones315, line 1518: Please be clear that this can take in either a single leaf, as a string, or a list of leaves.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1503,38 +1503,43 @@ def add_leaf(
...
 
         Raises:
-            CassiopeiaTreeError if the tree is not initialized or input node is
-            not a leaf
+            CassiopeiaTreeError if the tree is not initialized or any of the
+            input nodes are not leaves
```
mattjones315, line 1522: I think you need to indent the second line of the `Raises` statement for documentation generation.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1503,38 +1503,43 @@ def add_leaf(
...
         """
         self.__check_network_initialized()
 
-        if not self.is_leaf(node):
-            raise CassiopeiaTreeError("Node is not a leaf.")
+        if isinstance(nodes, str):
+            nodes = [nodes]
+        for n in nodes:
+            if not self.is_leaf(n):
+                raise CassiopeiaTreeError("A specified node is not a leaf.")
```
mattjones315, line 1530: Do you want to check this _before_ removing any leaves?

mattjones315 (changes requested):
Looks like a good fix! Left you a few comments to resolve.

mattjones315 (approved):
Looks good, thanks!

Issue #139, opened 2021-08-11
Deal with empty target site alleles appropriately


Lioscro:
Some target sites end up getting an empty string `''` as their allele. Still not entirely sure where/why this is happening, but the pipeline should deal with these more appropriately.
Currently, if a table has such empty-string-target site entries, reading the table with `pd.read_csv` causes these entries to be replaced with a floating-point `NaN`, which causes problems when we try to use the target site indels to construct a string. Specifically, if we try to join the target sites into a single allele string, as so
```python
molecule_table[['r1', 'r2', 'r3']].apply(lambda x: '_'.join(x))
```
this will cause an error because all the arguments to `join` must be strings, but `NaN` is a float.

One (hacky?) way to get around this is to simply provide `na_filter=False` to `pd.read_csv`, which reads the empty strings literally. But it is unclear if this is the behavior we want. Or do we want to simply filter out all UMIs with missing target site indels?

Here is an example of a UMI that has a missing target site indel.
```
AAAAAACCGTCGTA_GTGGTCGGG_1      777.0   110M1D52M1D55M  0.0     0.0     AATCCAGCTAGCTGTGCAGCCGTGAGTCTCTGATATTCAACTGCAGTAATGCTACCTAGTACTCACGCTTTCCAAGTGCTTGGCGTCGCATCACGGTCCTTTGTACGCCGAAAATCGCCAGACAACTAAGCTACGGCACGCTGCCATGTTGGGTCATACCCAAAATCAGGTTACTCCTGGGCCGCACAAGTCATGGAGAAGTCGAGACTATTAATTCATGATTCCCAATCTGGACTTACTACATGTATACCCC   GTGGTCGGG       [111:1D][164:1D]AAAAAACCGTCGTA  CGTGAGTCTCTGAT  [111:1D]        [164:1D]                1.0
```

After discussing with @mattjones315, this behavior is mainly due to the low sequencing quality of this dataset, compounded by technical artifacts in local alignment (no alignment for the particular target site)

Lioscro:
For the previous example, you could make the argument that there are a lot of mismatches at the third site, which is why the pipeline is having a hard time accurately calling the third indel (and hence the empty string). However, here is a more egregious example.
```
[...]

Issue #136, opened 2021-08-09
Missing data


oligomyeggo:
Hi again @mattjones315, I wanted follow up on the missing data issue that we discussed in issues #110 and #126. I went through the `allele_table` generated from my GESTALT data and counted up the number of instances I had a "Missing" entry/missing allele identity for each of my cut sites (r1, r2, r3, ... r10 for GESTALT). For cut sites r1-r8 there were no "Missing" entries. For cut site r9, ~24% of the entries were "Missing" and for cut site r10, ~36% of the entries were "Missing".  I know you mentioned [here](https://github.com/YosefLab/Cassiopeia/issues/110#issuecomment-889310054) that I can probably get away with missing data for now so long as it is less than ~30%, which is the case for r9 but not for r10. It seems interesting to me that the first 8 cut sites have no issues, and that missing data only increases at the end of my barcode array. Does this mean that in general the alignment is probably working, or do you think that we'll need to implement the global-alignment strategy described in the original GESTALT paper? Thanks!

mattjones315:
Hi @oligomyeggo ,

Apologies for not being prompt in my response here! We had some other issues that we noticed with missing data, especially when sequencing quality was low and were trying to come up with a best practice for this (see #139.)

The short answer is that we think that these are technical artifacts that should not be in the dataset. I think you might be losing meaningful information in r9 and r10 because of the local alignment strategy. We've made it a priority to try and fix this with a global alignment option but this is a bit tricky for some technical reasons (e.g., having to trim off the poly-A tail before global alignment). But in the meantime, we'd rather throw out cells with the missing information -- or, in your case, maybe you can just drop sites r9 and r10.

I would be curious to know if this is due to a sequencing quality issue. One way to try and assess this is to plot the indel heatmaps of trees (using, for example, `cas.pl.upload_and_export_itol`) -- does data look more "noisy" as you go from r1-r10?

Thanks and hope this helps. Sorry we don't have a solid fix for this yet!

oligomyeggo:
Hi @mattjones315, no worries at all! I never expect you (or any dev) to be able to respond to all my questions right away. I really just appreciate that you do respond, and that you've been so willing to help me process my GESTALT data, so thank you!

[...]

PR #142, opened 2021-08-13
global alignment

Lioscro:
Global alignment implementation (#127, #136, #139)
* Added a `method` argument to `pp.align_sequences` which can either be `local` (default) or `global`.
* Moved alignment functions to `pp.alignment_utilities`.
* Bumped some dependencies that were required.
* The gap open and gap extend penalties were left as is (20 and 1 respectively). Turns out this matches exactly to the values used in the original GESTALT paper (10 and 0.5; multiplicatively identical).

Regarding poly-A trimming, I'm not sure if this will be a big issue, but worth discussion.


cassiopeia/preprocess/alignment_utilities.py
```diff
@@ -3,10 +3,101 @@
...
+
+
+def align_global(
+    ref: str,
+    seq: str,
+    substitution_matrix: Dict[str, Dict[str, int]],
+    gap_open_penalty: int,
+    gap_extend_penalty: int,
+) -> Tuple[str, int, int, float, str]:
+    """Perform local alignment of `seq` to `ref` using Needleman-Wunsch.
```
mattjones315, line 73: This actually performs **global alignment** using Needleman-Wunsch.

cassiopeia/preprocess/alignment_utilities.py
```diff
@@ -3,10 +3,101 @@
...
+    Args:
+        ref: The reference sequence.
+        seq: The query sequence.
+        substitution_matrix: Nested dictionary encoding the substitution matrix.
+        gap_open_penalty: Gap open penalty.
+        gap_extend_penalty: Gap extend penalty.
+
+    Returns:
+        A tuple containing the CIGAR string, query sequence start position,
+        reference sequence start position, alignment score, and query sequence
```
mattjones315, line 84: Ident is necessary after the first line of the `Returns` statement.

cassiopeia/preprocess/alignment_utilities.py
```diff
@@ -3,10 +3,101 @@
...
```
mattjones315, line 40: Indent needed after first line of `Returns` statement.

cassiopeia/preprocess/pipeline.py
```diff
@@ -24,7 +24,7 @@
 from cassiopeia.mixins import logger, PreprocessError
 from cassiopeia.mixins.warnings import PreprocessWarning
 from cassiopeia.preprocess import (
-    alignment_utilities,
+    alignment_utilities as a_utils,
```
mattjones315, line 27: can we keep this as `alignment_utilities`? I find abbreviations like `a_utils` difficult to read, especially if I'm not familiar with the codebase.

Same as below for `map_utils` and `doublet_utils`.

cassiopeia/preprocess/pipeline.py
```diff
@@ -464,70 +464,71 @@ def align_sequences(
...
     Take in several queries stored in a DataFrame mapping cellBC-UMIs to a
-    sequence of interest and align each to a reference sequence. The alignment
-    algorithm used is the Smith-Waterman local alignment algorithm. The desired
-    output consists of the best alignment score and the CIGAR string storing the
-    indel locations in the query sequence.
+    sequence of interest and align each to a reference sequence. Either local
+    or global alignment may be performed, depending on the `method` argument.
+    The defaults for the gap open and gap extend penalties were selected via
+    in-silico simulation (which also happens to be equivalent to the values
+    used in the GESTALT technology described in McKenna et. al, 2016).
```
mattjones315, line 477: I'm not sure we should say this here -- because they differ in absolute terms.

mattjones315 (changes requested):
This is awesome! Left a few very minor comments.

We should really think about how to trim the poly-A tails, but let's save that for a future PR.

cassiopeia/preprocess/pipeline.py
Lioscro: How about now?

cassiopeia/preprocess/pipeline.py
```diff
@@ -464,70 +464,71 @@ def align_sequences(
...
     Take in several queries stored in a DataFrame mapping cellBC-UMIs to a
-    sequence of interest and align each to a reference sequence. The alignment
-    algorithm used is the Smith-Waterman local alignment algorithm. The desired
-    output consists of the best alignment score and the CIGAR string storing the
-    indel locations in the query sequence.
+    sequence of interest and align each to a reference sequence. Either local
+    or global alignment may be performed, depending on the `method` argument.
+    The defaults for the gap open and gap extend penalties were selected via
+    in-silico simulation (and are also equivalent to the values used in the
+    GESTALT technology described in McKenna et. al, 2016). The desired output
```
mattjones315, line 477: Sorry, I should have been more specific with what we should say. Since what you're trying to say is that the _ratio_ is the same and thus equivalent, why not say "and are functionally equivalent to the values used in the GESTALT technology described in McKenna et al, 2016"? Or something along those lines that discriminates between equal in absolute and relative terms.

cassiopeia/preprocess/pipeline.py
```diff
@@ -464,70 +464,71 @@ def align_sequences(
...
+        gap_open_penalty: Gap open penalty
+        gap_extend_penalty: Gap extension penalty
+        method: What alignment algorithm to use. Can be either "local" to
+            perform local alignment using Smith-Waterman or "global" to
+            perform global alignment using Needleman Wunsch.
         n_threads: Number of threads to use.
 
     Returns:
         A DataFrame mapping each sequence name to the CIGAR string, quality,
         and original query sequence.
```
mattjones315, line 494: Can you add an indent to the second line of this `Returns` statement?

mattjones315 (changes requested):
Love it! Two nitpicky details that I'd like to resolve before merging in. Thanks!

mattjones315 (approved):
Great! This looks like it's ready to be merged in. Thanks for the quick & effective work here!

Issue #138, opened 2021-08-10
Cleanup pandas operations in preprocessing pipeline


Lioscro:
The preprocessing pipeline performs a lot of grouped iterations of the following form:
```
for cellBC, group in moleculetable.groupby("cellBC"):
    ...
```
Wherever possible, we should update these instances to use groupby operations, instead of iterating over each group for the following reasons.
* Pandas groupby operations are usually much faster than iterating over each group and performing operations separately for each group.
* The code usually becomes inflated when iterating through each group manually, and thus increases code complexity.

This applies to several functions in `pp.utilities`, `pp.UMI_utils`, `pp.lineage_utils`, `pp.doublet_utils`.

PR #143, opened 2021-08-17
Preprocessing speedup

Lioscro:
#138 
Optimized the following functions 

| function                             | before       | after  | speedup |
|--------------------------------------|--------------|--------|---------|
| `pp.utilities.error_correct_intbc`   | 2 min        | 1 min  | 2x      |
| `pp.utilities.record_stats`          | 20 sec       | 178 ms | 140x    |
| `pp.UMI_utils.correct_umis_in_group` | 1 min 25 sec | 138 ms | 615x    |
| `pp.align_sequences` (10000 sequences) | 30 sec | 26 sec |  1.15x |
| `pp.map_utils.map_intbcs` | 6 min 54 sec | 3.47 sec | 119x |
| `pp.doublet_utils.filter_intra_doublets` | 49 sec | 361 ms | 135x |
| `pp.doublet_utils.get_intbc_set(s)` | 713 ms | 149 ms | 4x |
| `pp.doublet_utils.filter_inter_doublets` | 5 sec |  860 ms | 5x |

Other changes
* All instances of `moleculetable` have been made consistent to `molecule_table`
* Changed parameter names for `pp.utilities.error_correct_intbc` to snake-case.
* `pp.utilities.convert_bam_to_df` always returns a pandas Dataframe, and the caller must decide to write or not
* `pp.UMI_utils.correct_umis_in_group` no longer returns the total number of UMIs because that can be obtained by simply taking the number of rows in the returned dataframe.
* `pp.map_utils` is no longer a Cython extension, but a regular .py file. It was not using any Cython features.
* Changed `pp.doublet_utils.get_intbc_sets` to `pp.doublet_utils.get_intbc_set` such that the function operates on a single lineage group.
* `pp.doublet_utils` is no longer a Cython extension. It was not using any Cython features.


Lioscro:
Also @mattjones315, I've noticed in `pp.map_utils.map_intbcs`, the intBC with the highest *read count* is selected. Shouldn't we be doing this by UMI?

cassiopeia/preprocess/UMI_utils.py
```diff
@@ -659,61 +659,40 @@ def correct_umis_in_group(
...
-        input_df: Input DataFrame of alignments.
+        cell_group: Input DataFrame of alignments.
         max_umi_distance: Maximum Hamming distance between UMIs
             for error correction.
 
     Returns:
-        A DataFrame of error corrected UMIs within the grouping.
-    """
+        A DataFrame of error corrected UMIs within the grouping, and the
+        number of corrected UMIs
```
mattjones315, line 668: Need an extra indentation in any line after the first in `Returns` statements. Here and elsewhere.

cassiopeia/preprocess/UMI_utils.py
```diff
@@ -659,61 +659,40 @@ def correct_umis_in_group(
...
-
-        # Add alignments not touched during error correction back into the group
-        # to be written to file.
-        if al["UMI"] not in corrected_names:
-            corrected_group = corrected_group.append(al)
-
-    total = len(cell_group)
-
-    return corrected_group, num_corrections, total
+    # Perform correction by simple replacement
```
mattjones315, line 680: Wow. This is a _lot_ better!

cassiopeia/preprocess/map_utils.py
```diff
@@ -0,0 +1,54 @@
...
+"""
+
+import pandas as pd
+
+from cassiopeia.mixins import logger
+from cassiopeia.preprocess import utilities
+
+
+@utilities.log_molecule_table
+def map_intbcs(molecule_table: pd.DataFrame) -> pd.DataFrame:
```
mattjones315, line 13: You're right, I think we want to preferentially take the intBC with higher UMI count first, and then split by read count.

test/preprocess_tests/filter_molecule_table_test.py
```diff
@@ -7,7 +7,7 @@
 from cassiopeia.preprocess import pipeline
 
 
-class TestFilterMoleculeTable(unittest.TestCase):
+class TestFiltermolecule_table(unittest.TestCase):
```
mattjones315, line 10: Not a big deal, but normally class names are camel case in Python (at least in our codebase) so we can keep the old nomenclature.

mattjones315 (changes requested):
Looking great! Thanks for speeding up all of this code.

I agree, we should be mapping intBCs preferentially on the basis of #UMIs, not the read count.

test/preprocess_tests/filter_molecule_table_test.py
Lioscro: Whoops haha. It must have gotten replaced when I was replacing everything 😢

cassiopeia/preprocess/UMI_utils.py
Lioscro: Oh man I keep forgetting to do this. My bad.

mattjones315 (approved):
Looks good to me! No remaining comments here. Just make sure to merge in master before pushing.

PR #145, opened 2021-08-28
Parsimony and likelihood

richardyz98:
Hi Matt,
Here is the likelihood and parsimony PR.

In the likelihood, I use some marginalization tricks that are hopefully well explained in documentation. Additionally, please let me know if you disagree with any of the design choices.

Thanks so much,
Richard


cassiopeia/data/CassiopeiaTree.py
```diff
@@ -2046,6 +2061,315 @@ def get_unmutated_characters_along_edge(
...
+            for char in range(self.n_character):
+                likelihood_at_root = 0
+                for s_ in likelihoods_at_nodes[self.root][char]:
+                    likelihood_at_root += (
+                        transition_probability(0, s_, 1, char)
+                        * likelihoods_at_nodes[self.root][char][s_]
+                    )
+                likelihoods_at_implicit_root[char][0] = likelihood_at_root
+
+            return np.log(np.prod([i[0] for i in likelihoods_at_implicit_root]))
```
mattjones315, line 2361: why not use `np.sum([np.log(i[0]) for i in likelihoods_at_implicit_root])`? I think this will be more stable.

Also, let's be clear in the description that we're returning the _log_ likelihood.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -2046,6 +2061,315 @@ def get_unmutated_characters_along_edge(
...
+
+        else:
+            return np.log(
```
mattjones315, line 2364: ditto -- let's not compute a product and log it, let's compute the sum of logs.

and make it clear that we are returning the log likelihood.

mattjones315 (changes requested):
Thanks Richard!

The parsimony module looks great. 

As for the likelihood -- this looks good but there's some work to be done to make this more clear, and I think there are some areas I think might not be implementing the likelihood correctly as I point out. Generally I think that this function is a bit too complex for right now. We can likely make this a bit more simple, right?

richardyz98:
Hi Matt, 
I made some updates here. I added the naive estimation of rates, as well as some documentation fixes. 
Additionally, I thought that if we are to add an implicit edge and we're using branch lengths, we should probably make its branch length the average of the branch lengths, so I implemented that as well.
Finally, I added some additional tests around the estimation of rates.

Thanks!

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -686,6 +687,9 @@ def __remove_node(self, node) -> None:
 
         self.__network.remove_node(node)
 
+        # this will change the topology of the tree, so reset the cache
+        self.__cache = {}
```
sprillo, line 691: Is this a bugfix?

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -2046,6 +2061,564 @@ def get_unmutated_characters_along_edge(
...
+            else:
+                for char in range(self.n_character):
+                    child_possible_states = []
+                    for c in [
+                        set(likelihoods_at_nodes[child][char])
+                        for child in self.children(n)
+                    ]:
+                        if (
+                            self.missing_state_indicator not in c
+                            and "*" not in c
```
sprillo, line 2527: Why do we need the concept of `"*"` in the implementation? It makes things a lot harder to understand for me. Can we avoid it completely?

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -2046,6 +2061,564 @@ def get_unmutated_characters_along_edge(
...
+                                        )
+                                    ) * stochastic_missing_rate
+                                if s_ != -1:
+                                    likelihood_at_state_for_child *= (
+                                        1 - stochastic_missing_rate
+                                    )
+                            likelihood_for_child += (
+                                likelihood_at_state_for_child
+                            )
+                        likelihood_s *= likelihood_for_child
```
sprillo, line 2583: Work with log-likelihoods instead, and use `scipy.special.logsumexp` to perform marginalization.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -2046,6 +2061,564 @@ def get_unmutated_characters_along_edge(
...
+
+        # If we are not to use the internal state annotations explicitly,
+        # then we impose an implicit root of all 0s if it does not exist
+        if (
+            not use_internal_character_states
+            and len(self.children(self.root)) != 1
+        ):
+            # If branch lengths are to be used, we calculate the length of the
+            # implicit root edge as the average of all other edges
+            if use_branch_lengths:
```
sprillo, line 2596: The code has a huge amount of conditionals, which makes it really hard to verify correctness by reading it. Conditionals are probably not the right way to handle switching between the cases of `use_branch_lengths`. We should think of a better design.

sprillo (changes requested):
Thanks for all the hard work @richardyz98 ! This is a complex method to implement. All the pieces are there, but the code is just too complicated. There are too many conditionals resulting from the desire to implement many different behaviors (`use_branch_lengths`, `use_internal_character_states`) in the same method, as well as extra bells and whistles (e.g. reconstruction of ancestral states, inference of rates if not provided) which makes the code really long, branching, hard to read and test, and vulnerable to shotgun surgery.

@richardyz98 @mattjones315 we should think of a better way of factorizing out the implementation of likelihood computation for the different use cases (`use_branch_lengths`, `use_internal_character_states`). As with all things, it is probably best to start small. Lets implement and test just *one* specific case (note: it's fine if this implementation is informed by the other use cases we will implement later). Then, we can think of the new use case, figure out what should be abstracted out, refactor, and move on. In the end, this will take less time than trying to do everything at once. Incrementality/simplicity are key.

richardyz98:
Hi all,
I have thoroughly refactored the likelihood calculation. Hopefully most of your comments are addressed. Updates include:

- Moved the parameter estimation to a new `parameter_estimators` file in `tools`
- Refactored out the estimators into separate functions 
- Moved the parsimony and likelihood to a new `tree_metrics` file in `tools`
- Refactored likelihood in terms of performing likelihood per character, as well as refactored in terms of log likelihoods to improve numerical stability
- Updated documentation
- Updated testing

In regards to the 2 outstanding comments that Sebastian left:

- I gave this some thought and I do believe the use of "*" (now "&") is warranted. This marginalization trick makes it such that we do not have to marginalize over the entire state space in the case where all daughters of an internal node have the missing state while keeping the general structure of the code intact, cleaning up the code by avoiding having to add additional conditionals and such to account for this case
- In regards to decoupling the reconstruction of internal states, it could be helpful to have the option to do the reconstruction built into the functions that depend on them just for the sake of a user who may not understand that this is an option. I guess we can include in an error statement that this is an option when internal states are not annotated or throw this into the docstring for functions that depend on annotating internal states, but I think an argument is more explicit. @mattjones315 @sprillo let me know how you feel about this, and if you feel we should do the decoupling, then I can address that in a separate PR.

Thanks in advance! 
Richard

sprillo:
Thanks @richardyz98 ! I am currently swamped with other stuff (in particular reproducing results from the KP manuscript), so @mattjones315 can take care of this review. I am sure you have carefully considered my points.

test/tools_tests/tree_metrics_test.py
```diff
@@ -0,0 +1,414 @@
...
+        with self.assertRaises(CassiopeiaTreeError):
+            tm.calculate_parsimony(small_tree, infer_ancestral_characters=False)
+
+        small_tree.set_character_states("node7", [0, 0, 0])
+        small_tree.set_character_states("node5", [0, 0, 0])
+        with self.assertRaises(CassiopeiaTreeError):
+            tm.calculate_parsimony(small_tree, infer_ancestral_characters=False)
+
+    def test_parsimony_reconstruct_internal_states(self):
+        small_cm = pd.DataFrame.from_dict(
```
mattjones315, line 63: it appears that this character matrix is used several times throughout the tests, it would be a lot cleaner if you made this a part of the test class in the `setUp()` function.

mattjones315 (changes requested):
Hi @richardyz98 ,

Thanks for the hard work here! I'm afraid there's still quite some work to be done. There are lots of clerical errors in these methods that I've made note of that we must fix before pushing this in.

I also have a few comments on the structure / logic of the methods in `tree_metrics.py`.  Mainly they are around readability and complexity. I still feel that this method is a bit complex and lots of the logical flow can be simplified. Please let me know if these make sense. 

Looking forward to the next revision,
Matt

cassiopeia/tools/parameter_estimators.py
```diff
@@ -0,0 +1,426 @@
+"""
+This file stores functions for estimating lineage tracing parameters.
+Currently, we'll support the estimation of mutation and missing data rates.
+"""
+from cassiopeia.data.CassiopeiaTree import CassiopeiaTree
```
mattjones315, line 5: This import (`cassiopeia.data.CassiopeiaTree`) needs to go down with the other Cassiopeia import.

cassiopeia/tools/parameter_estimators.py
```diff
@@ -0,0 +1,426 @@
...
+    assume_root_implicit_branch: bool = True,
+    stochastic_missing_probability=None,
+    heritable_missing_rate=None,
+    layer: Optional[str] = None,
+) -> Tuple[float, float]:
+    """
+    Estimates both missing data parameters given one of the two from a tree.
+
+    The stochastic missing probability is the probability that any given
+    cell/character pair acquires stochastic missing data in the character
```
mattjones315, line 197: Can you briefly describe what the mechanism is behind "stochastic" missing data?

cassiopeia/tools/parameter_estimators.py
```diff
@@ -0,0 +1,426 @@
...
+        mutation_rate = -np.log(1 - mutation_proportion) / mean_time
+    return mutation_rate
+
+
+def estimate_missing_data_rates(
+    tree: CassiopeiaTree,
+    continuous: bool = True,
+    assume_root_implicit_branch: bool = True,
+    stochastic_missing_probability=None,
+    heritable_missing_rate=None,
```
mattjones315, line 190: these need types

cassiopeia/tools/parameter_estimators.py
```diff
@@ -0,0 +1,426 @@
...
+
+    Args:
+        tree: The CassiopeiaTree specifying the tree and the character matrix
+        continuous: Whether to calculate a continuous missing rate,
+            accounting for branch lengths. Otherwise, calculates a
+            discrete missing rate based on the number of generations
+        assume_root_implicit_branch: Whether to assume that there is an
+            implicit branch leading from the root, if it doesn't exist
+        stochastic_missing_probability: The stochastic missing probability.
+            Will override the value on the tree
```
mattjones315, line 268: Can you give people an idea of what a reasonable `stochastic_missing_probability` is? Say 10%?

mattjones315 (changes requested):
Great job @richardyz98! I think this is a lot cleaner and very clear code. I left two or three small comments that I think we should address before merging in.

Also before merging in, please make sure to add relevant functions to the tools/__init__.py function. I'll add functions to the documentation.

mattjones315 (approved):
Looks great! Let's merge this in.

sprillo (approved):
Thanks for taking care of this!

PR #147, opened 2021-09-21
pass read header to cluster function

Lioscro:
Fixes issue that occurs when trying to use the new pipeline with a BAM file generated via cellranger (a.k.a. old pipeline).


mattjones315 (approved):
Looks good! Thanks.

PR #148, opened 2021-10-20
Morans I

mattjones315:
Implemented the Moran's I statistic in cassiopeia/tools/autocorrelation.py.

To do so, I also implemented a helper function for getting phylogenetic weight matrices.

I also changed the behavior in initializing `CassiopeiaTrees` where if a networkx object has an attribute `length`, then the branch lengths will be initialized. We'll need to change the `ete3_to_networkx` function to also take this into account which can be a future enhancement.


mattjones315:
Thanks for the suggestions @Lioscro -- I think they improved this code a lot. Happy to integrate more of your suggestions after you have a chance to take a look again.

Issue #149, opened 2021-10-25
Implement Utilities from KP-Tracer MS


mattjones315:
Implement utilities from recent [KP-Tracer manuscript](https://www.biorxiv.org/content/10.1101/2021.10.12.464111v1) into the Cassiopeia codebase. Specifically:

- Expansion index
- Phylogenetic coupling
- Phylotime

PR #154, opened 2021-10-27
Evolutionary coupling

mattjones315:
Implements the Evolutionary Coupling statistic.

The main logic is stored in `cassiopeia/tools/coupling.py` which in the future can store other methods for computing the "couplings" between variables. 

Auxiliary functionality is implemented in `cassiopeia/data/utilities.py` for computing distances between categories.

This will work towards resolving #149


cassiopeia/data/utilities.py
```diff
@@ -412,3 +413,97 @@ def compute_phylogenetic_weight_matrix(
     np.fill_diagonal(W.values, 0)
 
     return W
+    
+@numba.jit(nopython=True)
+def net_relatedness_index(
+    dissimilarity_map: np.array, indices_1: np.array, indices_2: np.array
+) -> float: # pragma: no cover
```
Lioscro, line 420: What does this comment do?

cassiopeia/tools/coupling.py
```diff
@@ -0,0 +1,125 @@
...
+def compute_evolutionary_coupling(
+    tree: CassiopeiaTree,
+    meta_variable: str,
+    minimum_proportion: float = 0.05,
+    number_of_shuffles: int = 500,
+    random_state: Optional[np.random.RandomState] = None,
+    dissimilarity_map: Optional[pd.DataFrame] = None,
+    cluster_comparison_function: Optional[
+        Callable
+    ] = data_utilities.net_relatedness_index,
```
Lioscro, line 25: Think this is a typo. `Optional` is only used for arguments that are `None` by default.

Lioscro (changes requested):
Looks great! Just a couple very minor comments.

cassiopeia/data/utilities.py
mattjones315: I was thinking that it would ignore this function from the coverage report but I don't think it's the case. I'll get rid of it.

(The reason why I wanted to ignore it was apparently our coverage reports actually can't test functions wrapped in `numba`, so we have an artificially deflated coverage. Not a huge deal but something I was trying to counteract.)

cassiopeia/tools/coupling.py
mattjones315: You're totally right! I'll get rid of this.

Lioscro (approved):
LGTM!

PR #158, opened 2021-11-04
SNJ Implementation Draft 

ivalexander13:
**Summary**
Spectral Neighbor Joining algorithm (Jaffe 2020) implemented as a solver that implements DistanceSolver.

Also, an updated DistanceSolver method (get_dissimilarity_map) which replaces the same functionality previously done within solve(). This is to allow the use of alternative matrices besides the dissimilarity map in solve.

**Todos:**
- [ ] Add integration tests
- [ ] Ensure Black formatting and other stylistic guidelines

**Questions**
- What does add_root do in neighborjoining? It triggers setup root finder at the start (get dissimilarity map method), which adds another node called "root". What is this for? What about root_tree that's run after solve? Why root twice?
- What do we do with duplicate rows in a character matrix?

### DIFF
--- a/cassiopeia/solver/DistanceSolver.py
+++ b/cassiopeia/solver/DistanceSolver.py
@@ -75,7 +75,6 @@ class DistanceSolver(CassiopeiaSolver.CassiopeiaSolver):
         self.dissimilarity_function = dissimilarity_function
         self.add_root = add_root
 
-    # @override
     def get_dissimilarity_map(
         self, 
         cassiopeia_tree: CassiopeiaTree,
--- a/cassiopeia/solver/DistanceSolver.py
+++ b/cassiopeia/solver/DistanceSolver.py
@@ -75,6 +75,19 @@ class DistanceSolver(CassiopeiaSolver.CassiopeiaSolver):
         self.dissimilarity_function = dissimilarity_function
         self.add_root = add_root
 
+    # @override
+    def get_dissimilarity_map(
+        self, 
+        cassiopeia_tree: CassiopeiaTree,
+        layer: Optional[str] = None
+    ) -> pd.DataFrame: #todo: docstring
+
+        self.setup_dissimilarity_map(cassiopeia_tree, layer)
+        dissimilarity_map = cassiopeia_tree.get_dissimilarity_map()
+
+        return dissimilarity_map
+
+
     def solve(
         self,
         cassiopeia_tree: CassiopeiaTree,
@@ -103,9 +116,7 @@ class DistanceSolver(CassiopeiaSolver.CassiopeiaSolver):
         """
         node_name_generator = solver_utilities.node_name_generator()
 
-        self.setup_dissimilarity_map(cassiopeia_tree, layer)
-
-        dissimilarity_map = cassiopeia_tree.get_dissimilarity_map()
+        dissimilarity_map = self.get_dissimilarity_map(cassiopeia_tree, layer)
 
         N = dissimilarity_map.shape[0]
 
--- a/cassiopeia/solver/SpectralNeighborJoiningSolver.py
+++ b/cassiopeia/solver/SpectralNeighborJoiningSolver.py
@@ -10,15 +10,21 @@ import numpy as np
 import pandas as pd
 from itertools import chain
 
+# Only for debugging purposes
+if True:
+    import sys
+    sys.path.insert(0, '../..')
+
+from cassiopeia.solver.DistanceSolver import DistanceSolver
+
 from cassiopeia.data import CassiopeiaTree
 from cassiopeia.solver import (
-    NeighborJoiningSolver,
     dissimilarity_functions,
     solver_utilities,
 )
 
 
-class SpectralNeighborJoiningSolver(NeighborJoiningSolver):
+class SpectralNeighborJoiningSolver(DistanceSolver):
     """
     Spectral Neighbor-joining class for Cassiopeia.
 
@@ -40,30 +46,81 @@ class SpectralNeighborJoiningSolver(NeighborJoiningSolver):
     ):
         super().__init__(dissimilarity_function, add_root=add_root, prior_transformation=prior_transformation) #type: ignore
 
+    def root_tree(
+        self, tree: nx.Graph, root_sample: str, remaining_samples: List[str]
+    ) -> nx.DiGraph():
+        """Roots a tree produced by Neighbor-Joining at the specified root.
 
-    def find_cherry(self, dissimilarity_matrix: np.ndarray, lambda_indices: List[List[int]] = None) -> Tuple[int, int]:
-        """Finds a pair of samples to join into a cherry.
-
-        Proceeds by minimizing the second largest SVD of the RA Matrix of some subset of nodes.
+        Uses the specified root to root the tree passed in
 
         Args:
-            dissimilarity_matrix: A sample x sample dissimilarity matrix
+            tree: Networkx object representing the tree topology
+            root_sample: Sample to treat as the root
+            remaining_samples: The last two unjoined nodes in the tree
+
+        Returns:
+            A rooted tree
+        """
+        tree.add_edge(remaining_samples[0], remaining_samples[1])
+
+        rooted_tree = nx.DiGraph()
+        for e in nx.dfs_edges(tree, source=root_sample):
+            rooted_tree.add_edge(e[0], e[1])
+
+        return rooted_tree
+
+    def get_dissimilarity_map(
+        self, 
+        cassiopeia_tree: CassiopeiaTree,
+        layer: Optional[str] = None
+    ) -> pd.DataFrame: 
+        """Outputs the first lambda matrix, where every subset is an individual node.
+
+        Args:
+            cassiopeia_tree: the CassiopeiaTree object passed into solve()
+            layer: Layer storing the character matrix for solving. If None, the
+                default character matrix is used in the CassiopeiaTree.
+
+        Returns:
+            DataFrame object of the lamba matrix, but similar in structure to DistanceSolver's dissimilarity_map.
+        """
+
+        # get similarity map and save it as private instance variable
+        self.__similarity_map = super().get_dissimilarity_map(cassiopeia_tree, layer)
+
+        # generate first (pairwise) lambda matrix
+        N = self.__similarity_map.shape[0]
+        self.lambda_indices: List[List[int]] = [[i] for i in range(N)]
+        node_names: np.ndarray = self.__similarity_map.index.values
+
+        lambda_matrix_arr = self.__compute_lambda(self.lambda_indices)
+
+        lambda_matrix_df = pd.DataFrame(
+            lambda_matrix_arr, 
+            index=node_names,
+            columns=node_names
+        )        
+
+        return lambda_matrix_df
+
+
+    def find_cherry(self, dissimilarity_map: np.ndarray) -> Tuple[int, int]:
+        """Finds a pair of samples to join into a cherry.
+
+        With dissimilarity_map being the lambda matrix, this method finds the argmin pair of subsets of the lambda matrix. 
+
+        Args:
+            dissimilarity_matrix: Lambda matrix
 
         Returns:
             A tuple of intgers representing rows in the dissimilarity matrix
                 to join.
         """
-        # temp? to pass tests
-        if not lambda_indices:
-            lambda_indices = [[i] for i in range(dissimilarity_matrix.shape[0])]
 
-        lambda_matrix = self.__compute_lambda(dissimilarity_matrix, lambda_indices)
-        np.fill_diagonal(lambda_matrix, np.inf)
-
-        return np.unravel_index(np.argmin(lambda_matrix, axis=None), lambda_matrix.shape) # TODO: cache with heapq?
+        return np.unravel_index(np.argmin(dissimilarity_map, axis=None), dissimilarity_map.shape) # type: ignore
 
 
-    def __compute_lambda(self, dissimilarity_map: np.ndarray, lambda_indices: List[List[int]]) -> np.ndarray:
+    def __compute_lambda(self, lambda_indices: List[List[int]]) -> np.ndarray:
         """Computes the lambda matrix, whose entries consist of the second SVD of the R^A affinity matrix for every pair of nodes.
 
         Computes the lambda(i,j) = SVD # 2 for R^(Ai U Aj) where i and j are the susbet indices in lambda_indices.
@@ -76,7 +133,7 @@ class SpectralNeighborJoiningSolver(NeighborJoiningSolver):
         """
 
         n = len(lambda_indices)
-        lambda_matrix = np.zeros([n, n])
+        lambda_matrix_arr = np.zeros([n, n])
 
         for i_count, i in enumerate(lambda_indices):
             for j_count, j in enumerate(lambda_indices):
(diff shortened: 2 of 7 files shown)


mattjones315:
Thanks for posting this @ivalexander13 and @kevinan1! 

I can answer your two questions right now before taking a look at the PR -- 

(1) The `add_root` method in the DistanceSolver handles the scenario in which an algorithm returns an unrooted tree and we would like to convert it to a rooted one. There is logic for doing this stored in `NeighborJoining`. `UPGMA` returns a rooted tree, so this is a no-op. SNJ I believe does not return a rooted tree, so you can use the same logic as in the NJ solver. You can also read more about the functionality of the `add_root` argument [here](https://cassiopeia-lineage.readthedocs.io/en/latest/api/reference/cassiopeia.solver.NeighborJoiningSolver.html#cassiopeia.solver.NeighborJoiningSolver) and [here](https://github.com/YosefLab/Cassiopeia/blob/d13d3e72acbeabd807af01356acb7a9783346aca/cassiopeia/solver/NeighborJoiningSolver.py#L73).

(2) Currently we are building the tree over all samples, even if they are duplicated character states. There's a future endeavor we have in mind to speed things up by removing duplicates. This won't change the asymptotic performance, however, and you can proceed by building over all samples.

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
+"""
+This file stores a subclass of DistanceSolver, SpectralNeighborJoining. This algorithm is based on the one developed by Jaffe, et al. (2020) in their paper titled "Spectral neighbor joining for reconstruction of latent tree models.
```
mattjones315, line 2: Please clean up formatting in this head docstring - you're missing an end quote after the title of the paper. Also, please add the journal name after the reference.

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+This file stores a subclass of DistanceSolver, SpectralNeighborJoining. This algorithm is based on the one developed by Jaffe, et al. (2020) in their paper titled "Spectral neighbor joining for reconstruction of latent tree models. 
+"""
+
+from typing import Callable, Dict, List, Optional, Tuple, Union
+
+import abc
+import networkx as nx
+import numpy as np
+import pandas as pd
+from itertools import chain
```
mattjones315, line 11: alphabetize your imports

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+
+# Only for debugging purposes
+# if True:
+#     import sys
+#     sys.path.insert(0, '../..')
+
```
mattjones315, line 17: please remove this when you submit your PR for final review.

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+from cassiopeia.data import CassiopeiaTree
+from cassiopeia.solver import (
+    dissimilarity_functions,
+    solver_utilities,
+)
+
+from sklearn.decomposition import TruncatedSVD
+from scipy.sparse.linalg import svds, eigs
+from scipy.linalg import svd as scipy_svd
```
mattjones315, line 28: Alphabetize, and place these imports above with the other third-party imports (e.g., `pandas`).

You also have three different SVD algorithms. Only import and use one.

Also, you will have to add `sklearn` and `scipy` as dependencies in the `setup.py` file for the builds to pass.

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+class SpectralNeighborJoiningSolver(DistanceSolver):
+    """
+    Spectral Neighbor-joining class for Cassiopeia.
+
+    Implements a variation on the Spectral Neighbor-Joining algorithm described by Jaffe et al. in their 2020 paper. This class implements the abstract class DistanceSolver. This class inherits the 'solve' method from DistanceSolver with the find_cherry method implemented using a spectral method as in Jaffe et al. (2020).
+
+    Args:
+        dissimilarity_function: A function by which to compute the dissimilarity map, corresponding to the affinity R(i, j) in Jaffe et al. (2020). By default we will use exp(-d(i,j)) where d is the metric given by weighted_hamming_distance from dissimilarity functions without weights given.
```
mattjones315, line 38: Be sure all lines (including comments) are 80 characters maximum.

Also, you'll have to add more arguments beyond just dissimilarity function. Take a look at the formatting of other `Args` documentation in the other solvers.

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+        prior_transformation: str = "negative_log",
+    ):
+        super().__init__(dissimilarity_function, add_root=add_root, prior_transformation=prior_transformation) #type: ignore
+
+    def get_dissimilarity_map(
+        self, 
+        cassiopeia_tree: CassiopeiaTree,
+        layer: Optional[str] = None
+    ) -> pd.DataFrame: 
+        """Outputs the first lambda matrix, where every subset is an individual node.
```
mattjones315, line 58: This title line of the docstring must be a single line less than 80 characters long.

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+
+        Args:
+            cassiopeia_tree: the CassiopeiaTree object passed into solve()
+            layer: Layer storing the character matrix for solving. If None, the
+                default character matrix is used in the CassiopeiaTree.
+
+        Returns:
+            DataFrame object of the lamba matrix, but similar in structure to DistanceSolver's dissimilarity_map.
```
mattjones315, line 66: The second line of a `Returns` statement must be indented.

cassiopeia/solver/DistanceSolver.py
```diff
@@ -75,6 +75,19 @@ def __init__(
         self.dissimilarity_function = dissimilarity_function
         self.add_root = add_root
 
+    # @override
```
mattjones315, line 78: What does this pragma "# @override" mean / do?

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+        """
+
+        # get similarity map and save it as private instance variable
+        self.__similarity_map = super().get_dissimilarity_map(cassiopeia_tree, layer)
```
mattjones315, line 70: Why do you assume that the `dissimilarity_map` stored in the `CassiopeiaTree` is a similarity map and not a dissimilarity map? This is not clear from the documentation, and may account for some of the integration issues that @kevinan1 is observing?

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+        Returns:
+            A tuple of intgers representing rows in the dissimilarity matrix
+                to join.
+        """
+
+        return np.unravel_index(np.argmin(dissimilarity_map, axis=None), dissimilarity_map.shape)
+
+
+    def compute_lambda(self, lambda_indices: List[List[int]]) -> np.ndarray:
+        """Computes the lambda matrix, whose entries consist of the second SVD of the R^A affinity matrix for every pair of nodes.
```
mattjones315, line 105: ditto above - one line for the docstring header (fewer than 80 characters)

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+
+        Computes the lambda(i,j) = SVD # 2 for R^(Ai U Aj) where i and j are the susbet indices in lambda_indices.
```
mattjones315, line 107: SVD #2 is a bit confusing, maybe you want to say "second singular value" explicitly.

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+
+        Args:
+            dissimilarity_map: A sample x sample dissimilarity map
+
+        Returns:
+            A matrix storing the Q-criterion for every pair of samples.
```
mattjones315, line 113: This is not going to store the "Q-criterion" for every pair of samples, but rather the second singular value for every pair of clades.

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+        return np.unravel_index(np.argmin(dissimilarity_map, axis=None), dissimilarity_map.shape)
+
+
+    def compute_lambda(self, lambda_indices: List[List[int]]) -> np.ndarray:
+        """Computes the lambda matrix, whose entries consist of the second SVD of the R^A affinity matrix for every pair of nodes.
+
+        Computes the lambda(i,j) = SVD # 2 for R^(Ai U Aj) where i and j are the susbet indices in lambda_indices.
+
+        Args:
+            dissimilarity_map: A sample x sample dissimilarity map
```
mattjones315, line 110: Actually, this must be a `sample x samples` __**similarity**__ map. 

On another note, I don't think that this Arguments section reflects the arguments actually passed in.

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+        lambda_indices_copy.append([
+            *lambda_indices_copy[i], 
+            *lambda_indices_copy[j]
+            ])
+        lambda_indices_copy.pop(max(i, j))
+        lambda_indices_copy.pop(min(i, j))
+
+        if len(lambda_indices_copy) > 2:
+            # compute new lambda matrix
+            lambda_matrix_arr = self.compute_lambda(lambda_indices_copy)
```
mattjones315, line 194: This line seems like it is going to _recompute_ values that already have been computed. You really only need to add more column / row to your lambda matrix. Correct me if I'm wrong.

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+            columns=node_names
+        )
+
+        # update lambda indices
+        self.lambda_indices = lambda_indices_copy
+
+        return lambda_matrix_df
+
+    def setup_root_finder(self, cassiopeia_tree: CassiopeiaTree) -> None:
+        """Defines the implicit rooting strategy for the NeighborJoiningSolver.
```
mattjones315, line 212: typo: `NeighborJoiningSolver` -> `SpectralNeighborJoiningSolver`.

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+
+        By default, the SpectralNeighborJoining algorithm returns an unrooted tree.
```
mattjones315, line 214: Line length.

Also -- if this function is taken directly from `NeighborJoiningSolver`, make it clear that we are duplicating it here.

cassiopeia/solver/dissimilarity_functions.py
```diff
@@ -238,6 +238,34 @@ def weighted_hamming_similarity(
...
+    s1: List[int],
+    s2: List[int],
+    missing_state_indicator=-1,
+    weights: Optional[Dict[int, Dict[int, float]]] = None,
+) -> float:
+
+
+    """
+    Calculates the default affinity for SpectralNeighborJoiningSolver. This simply returns exp(-d(i,j)) where d is the weighted_hamming_distance function as above where no weights are passed in.
+    In other words, we increment dissimilarity score d by +2 if the states are different, +1 if one state is uncut and the other is an indel, and +0 if the two states are identical. Then, we take this total d and return exp(-d).
```
mattjones315, line 251: Line length.

cassiopeia/solver/dissimilarity_functions.py
```diff
@@ -238,6 +238,34 @@ def weighted_hamming_similarity(
...
+
+    Args:
+        s1: Character states of the first sample
+        s2: Character states of the second sample
+        missing_state_indicator: The character representing missing values
+        weights: A dictionary storing the state weights for each character, derived
```
mattjones315, line 257: line length.

cassiopeia/solver/dissimilarity_functions.py
```diff
@@ -238,6 +238,34 @@ def weighted_hamming_similarity(
...
+            from the state priors. This should be a nested dictionary where each
+            key corresponds to character that then indexes another dictionary
+            storing the weight of each observed state.
+            (Character -> State -> Weight)
+
+    Returns:
+        A dissimilarity score.
+    """
+
+    return 2.718281828459045 ** (- weighted_hamming_distance(s1, s2, missing_state_indicator=missing_state_indicator))
```
mattjones315, line 267: Why not use `np.exp` instead of a numerical estimation of `e`?

mattjones315 (commented):
Great job @ivalexander13 and @kevinan1 !

I've taken a pass through the draft and hopefully have left helpful comments. There are several "easy" comments to address - mostly around formatting and line length.

There are some logical problems that I believe I found and possibly flagged in your implementation, please let me know if anything is unclear or I am incorrect.

I'll be happy to take a deeper look at the tests when you have everything passing.

Thanks again and looking forward to your updates!

test/solver_tests/snj_solver_test.py
```diff
@@ -0,0 +1,642 @@
...
+
+    def test_compute_lambda_pairwise(self):
+        self.lambda_pairwise = [[i] for i in range(self.sim_general.values.shape[0])]
+
+        # run to generate similarity map
+        self.snj_solver.get_dissimilarity_map(self.tree_lineage)
+
+        svd2_vals = self.snj_solver.compute_lambda(self.lambda_pairwise)
+
+        expected_q = pd.DataFrame.from_dict(
```
richardyz98, line 200: Please standardize these variable names for similar outputs throughout. For example, here it's called `expected_q`, but later is called `expected_lambda` and `expected_lambda_matrix`.

richardyz98 (commented):
Hi Kevin and Ivan,
The implementation is looking great so far! Y'all have done some great work.
I know this is still currently a draft but I took the liberty of going through and adding some hopefully helpful comments. Let me know if you have any questions regarding my comments. 

One major thing that Matt and I have have both noted is that it seems in `compute_lambda`, you are recalculating the entire lambda matrix at each merging step. I think that runtime performance could be enhanced if we only calculated the new entries of the lambda matrix when we join a new cluster. 

Thanks again for the hard work!

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+        node_names = np.append(node_names, new_node)
+        node_names = np.delete(node_names, [i, j])
+
+        # modify indices
+        lambda_indices_copy.append([
+            *lambda_indices_copy[i], 
+            *lambda_indices_copy[j]
+            ])
+        lambda_indices_copy.pop(max(i, j))
+        lambda_indices_copy.pop(min(i, j))
```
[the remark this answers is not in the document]
ivalexander13: pop() seems like the simplest way to delete items. "del" seems to only work for consecutive items, and list comprehension seems too long. The min and max is to avoid the case where the smaller value is used first and the rest of the indices are shifted, which makes the second pop to mis-pop.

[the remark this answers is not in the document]
ivalexander13: actually thinking about it more, j is always strictly less than i, so i can do pop(i) and then pop(j) to avoid the same issue. Thanks!

cassiopeia/solver/SpectralNeighborJoiningSolver.py
kevinan1: Is there a specific format in which I should include the journal name or should I just say something like "in their paper... published in the SIAM Journal for Math and Data Science."? a brief google search didn't seem to give any standard formats for doing this in text.

(also I'm assuming this is intended but I'll mention it just in case it needs to be corrected, I just sorta copied the format of the blurb under NeighborJoining script and it doesn't mention a journal with the citation)

cassiopeia/solver/SpectralNeighborJoiningSolver.py
kevinan1: just the comment part, right? 
"Outputs the first lambda matrix, where every subset is an individual node."
is less than 80 chars

cassiopeia/solver/SpectralNeighborJoiningSolver.py
```diff
@@ -0,0 +1,281 @@
...
+from scipy.linalg import svd as scipy_svd
+
+
+class SpectralNeighborJoiningSolver(DistanceSolver):
+    """
+    Spectral Neighbor-joining class for Cassiopeia.
+
+    Implements a variation on the Spectral Neighbor-Joining algorithm described by Jaffe et al. in their 2020 paper. This class implements the abstract class DistanceSolver. This class inherits the 'solve' method from DistanceSolver with the find_cherry method implemented using a spectral method as in Jaffe et al. (2020).
+
+    Args:
```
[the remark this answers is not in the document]
kevinan1: so I put the default for dissimilarity_map ('exponential_negative_hamming_distance') as a method in the dissimilarity_functions.py

Would it be weird to put a similarity map in the dissimilarity_functions.py?

I was also thinking of just taking in dissimilarity maps and transforming it into a similarity one when we actually use it but this is probably a terrible idea...?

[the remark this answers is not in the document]
kevinan1: so I tried changing replacing the places where we use "dissimilarity" with similarity, but because the abstract class DistanceSolver uses methods that are called get_dissimilarity_map or update_dissimilarity_map, when i rename them to similarity it errors out. Is there a cleaner way to do this? should i rename the methods in the DistanceSolver class? somethign like "affinity" perhaps?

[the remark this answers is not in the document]
kevinan1: nvm we'd need to change a bunch of files if we did this

PR #159, opened 2021-11-12
local plotting

Lioscro:
Support for plotting trees locally, instead of submitting to iTOL (since this is paid now...)

- [x] Support rectangle plots in the four cardinal directions (left, right, up, down)
- [x] Support polar plots (circular plots)
- [x] Support colorstrip annotations (where categorical/continuous/indel information is displayed as a color strip)
- [x] Support subclade coloring
- [x] Static plotting using Matplotlib
- [x] Interactive plotting using Plotly
- [x] Tutorial notebook (https://nbviewer.org/github/YosefLab/Cassiopeia/blob/tree-plotter/notebooks/local_plotting.ipynb) -- Github doesn't render plotly plots when viewing notebook.

Samples (3730_NT_T1)

Polar
image

Rectangle
image


mattjones315:
I'm really excited to take a look at this @Lioscro! 

One quick comment -- I think this PR will really benefit from a little notebook tutorial. What do you think?

Lioscro:
> I'm really excited to take a look at this @Lioscro!
> 
> One quick comment -- I think this PR will really benefit from a little notebook tutorial. What do you think?

Of course! Keep in mind there are still a ton of things to add to this PR, so I'll get to that eventually.

cassiopeia/plotting/local.py
```diff
@@ -0,0 +1,830 @@
...
+        line_width=0,
+        fill="toself",
+        mode="lines",
+        showlegend=False,
+        hoverinfo="text",
+        hoveron="fills",
+    )
+    _colorstrip_kwargs.update(colorstrip_kwargs or {})
+    for colorstrip in colorstrips:
+        # Last element is text, but this can not be shown in static plotting.
```
richardyz98, line 811: comment no longer necessary for this version it seems

cassiopeia/plotting/local.py
```diff
@@ -0,0 +1,830 @@
...
+    """Generate a static plot of a tree using Matplotlib.
+
+    Args:
+        tree: The CassiopeiaTree to plot.
+        depth_key: The node attribute to use as the depth of the nodes. If
+            not provided, the distances from the root is used by calling
+            `tree.get_distances`.
+        meta_data: Meta data to plot alongside the tree, which must be columns
+            in the CassiopeiaTree.cell_meta variable.
+        allele_table: Alleletable to plot alongside the tree.
```
richardyz98, line 614: "Allele table"?

cassiopeia/plotting/local.py
```diff
@@ -0,0 +1,830 @@
...
```
richardyz98, line 431: "Allele table"?

richardyz98 (changes requested):
Hi Joseph,
Amazing PR! The code is really clean and understandable. Thanks for this tremendous effort.

I had some clerical comments, but otherwise it looks really good.

Also, if I understand correctly, the plotting functions at the moment can only consume allele tables? I don't think it's in the scope of this PR, but perhaps we can add (and this is something I may be able to help on) functionality to consume character matrices as well?

Thanks! 
Richard

cassiopeia/plotting/utilities.py
```diff
@@ -0,0 +1,525 @@
...
+        }
+    else:
+        depths = tree.get_distances(root)
+
+    placement_depths = {}
+    positions = {}
+    leaf_i = 0
+    leaves = set()
+    for node in tree.depth_first_traverse_nodes(postorder=False):
+        if tree.is_leaf(node):
```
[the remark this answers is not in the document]
Lioscro: I'm using DFS here on purpose, because I need a specific ordering of leaves, such that siblings are grouped together. This is necessary for plotting because otherwise there may be branches that overlap!

cassiopeia/plotting/local.py
```diff
@@ -0,0 +1,830 @@
...
+        indel_priors: Prior probabilities for each indel. Only useful if an
+            allele table is to be plotted and `indel_colors` is None.
+        orient: The orientation of the tree. Valid arguments are `left`, `right`,
+            `up`, `down` to display a rectangular plot (indicating the direction
+            of going from root -> leaves) or any number, in which case the
+            tree is placed in polar coordinates with the provided number used
+            as an angle offset. Defaults to 90.
+        extend_branches: Extend branch lengths such that the distance from the
+            root to every node is the same. If `depth_key` is also provided, then
+            only the leaf branches are extended to the deepest leaf.
```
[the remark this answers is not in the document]
Lioscro: This is a somewhat arbitrary decision on my part so that the `depth_key` and `extend_branches` arguments work nicely with each other. Also, I previously saw Sebastian extending only the leaves to the deepest leaf in his branch length estimation stuff, so that's why I decided to include this.

cassiopeia/plotting/local.py
Lioscro: You're right. I copied over the matplotlib function as a template, so that's where these stragglers are coming from.

Lioscro:
> Hi Joseph, Amazing PR! The code is really clean and understandable. Thanks for this tremendous effort.
> 
> I had some clerical comments, but otherwise it looks really good.
> 
> Also, if I understand correctly, the plotting functions at the moment can only consume allele tables? I don't think it's in the scope of this PR, but perhaps we can add (and this is something I may be able to help on) functionality to consume character matrices as well?
> 
> Thanks! Richard

I think that is a great idea! I'll open a new issue to implement this. I also considered doing that in this PR, but at the moment I just wanted to replicate all the important functionalities from iTOL plotting.

cassiopeia/plotting/local.py
```diff
@@ -0,0 +1,829 @@
+import warnings
```
mattjones315, line 1: Can you add a docstring at the top of this file describing the contents?

cassiopeia/plotting/local.py
```diff
@@ -0,0 +1,829 @@
...
+from collections import defaultdict
+from typing import Dict, List, Optional, Tuple, Union
+
+import matplotlib as mpl
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import plotly.graph_objects as go
+from matplotlib.colors import hsv_to_rgb
+from typing_extensions import Literal
```
mattjones315, line 11: should this go above with the other `typing` imports?

cassiopeia/plotting/local.py
```diff
@@ -0,0 +1,829 @@
...
+        max_pos = max(max_pos, pos)
+        min_depth = min(min_depth, depth)
+        max_depth = max(max_depth, depth)
+    return (
+        (max_depth - min_depth) * 0.05,
+        (max_pos - min_pos + 1) / len(anchor_coords),
+    )
+
+
+def create_categorical_colorstrip(
```
mattjones315, line 58: Do you think it would be worthwhile to support users passing in their own `value_mapping` dictionaries?

cassiopeia/plotting/local.py
```diff
@@ -0,0 +1,829 @@
...
+
+
+def create_continuous_colorstrip(
+    values: Dict[str, float],
+    anchor_coords: Dict[str, Tuple[float, float]],
+    width: float,
+    height: float,
+    spacing: float,
+    loc: Literal["left", "right", "up", "down", "polar"],
+    cmap: Union[str, mpl.colors.Colormap] = "viridis",
```
mattjones315, line 106: Is it worthwhile to support adding `vmin` or `vmax` parameters (perhaps percentiles, like `scanpy` supports?)

cassiopeia/plotting/local.py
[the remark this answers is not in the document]
mattjones315: I agree - this is probably the best way to do this.

cassiopeia/plotting/utilities.py
```diff
@@ -0,0 +1,525 @@
...
+
+    Returns:
+        A tuple of x and y coordinates
+    """
+    theta = degrees_to_radians(degrees)
+    return r * np.cos(theta), r * np.sin(theta)
+
+
+def polars_to_cartesians(
+    degrees: List[float], rs: List[float]
```
mattjones315, line 45: Would it make more sense to pass in a list of `Tuples`?

cassiopeia/plotting/utilities.py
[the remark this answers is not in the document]
mattjones315: I have another question -- how difficult would it be to "reorder" leaves such that some criteria is met? Like if you have a huge rake of leaves that have the same character state and you want to group together leaves that have the same cluster ID? 

It almost seems like it would be as easy as just grouping first by the attribute in the cell_meta variable and then outputting coordinates of leaves by category.

mattjones315 (changes requested):
Wow! I am thoroughly impressed by this fantastic PR! I don't have anything I'd like to see changed per se, but before you merge this in, can you add these functions to the documentation website? I'm happy to help if you're unfamiliar with how our `docs` folder works. For this reason, I'll Request Changes until we get the docs updated.

cassiopeia/plotting/local.py
Lioscro: I'm not sure which way we prefer. `typing_extensions` is technically a 3rd party library, so it should come after native Python libraries? This is also what is done in `itol_utilities.py`

cassiopeia/plotting/local.py
Lioscro: Yep, although I don't think its worth adding support for the percentile string...

cassiopeia/plotting/utilities.py
Lioscro: I think the way it is currently makes the function easier to call, since all the coordinates are currently grouped by degrees and radians as separate lists. If we were to change this to take a list of tuples, every time we call this function we would have to transform the list of degrees and list of radians to a single list of tuples. I'll just leave this as is for now.

cassiopeia/plotting/utilities.py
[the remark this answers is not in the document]
Lioscro: Do you mean reordering the leaves within a clade according to some variable? I think that would require quite a bit of bookkeeping to make sure that no branches overlap in the visualization.

richardyz98 (approved):
Looks good to me Joseph! Thank you for clearing up all of my comments!

cassiopeia/plotting/local.py
mattjones315: Yeah that sounds reasonable, thanks for clarifying!

cassiopeia/plotting/local.py
mattjones315: For sure, that's fine. I definitely make use of the percentile strings in `scanpy`, but perhaps I'm the only one...

mattjones315 (approved):
This looks fantastic! Thanks so much for the hard work here. The only thing missing is the notebook tutorial on the documentation site. I'll add that in here for you.

PR #162, opened 2022-01-14
change numpy requirement in pyproject.toml to use oldest-supported-numpy

Lioscro:
I've recently been working more with Cython extensions, and there can be some esoteric bugs related to using Numpy C++ extensions from Cython code when pip's build isolation is used (i.e. when `pyproject.toml` is present).
Specifically, the following error can occur when specifying Numpy within `pyproject.toml`'s `build-system` directive.
```
Traceback (most recent call last):
  File "/home/kmin/.local/bin/cassiopeia-preprocess", line 5, in <module>
    from cassiopeia.preprocess.cassiopeia_preprocess import main
  File "/home/kmin/.local/lib/python3.8/site-packages/cassiopeia/__init__.py", line 11, in <module>
    from . import solver
  File "/home/kmin/.local/lib/python3.8/site-packages/cassiopeia/solver/__init__.py", line 8, in <module>
    from .ILPSolver import ILPSolver
  File "/home/kmin/.local/lib/python3.8/site-packages/cassiopeia/solver/ILPSolver.py", line 21, in <module>
    from cassiopeia.solver import (
  File "cassiopeia/solver/ilp_solver_utilities.pyx", line 1, in init cassiopeia.solver.ilp_solver_utilities
ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```
Which is usually because when pip builds a package with `pyproject.toml` present, it creates a new sandboxed environment with the required build system packages, to which the package is built. Specifically, if Numpy is directly specified in `pyproject.toml`, the Numpy version used by the build system may differ from the actual Numpy version installed on the system, thus causing the package to be built against a Numpy version that is actually different from that which will be used when running the package.

Specifying `oldest-supported-numpy` instead of `numpy` directly seems to fix the issue (at least in all cases where I've experienced this problem).

Specific changes in this PR:
* `oldest-supported-numpy` used to specify numpy requirement in `pyproject.toml`, instead of using `numpy` directly.
* Changed the order of the commands in the `install` directive of the `Makefile`. `pip install` comes before `build`.
* Added `.gitignore` entries from Github's Python gitignore template.
* Added generated cpp files to gitignore.
* Modified `MANIFEST.in` to recursively include all files under the `cassiopeia` directory.
* Added BLE files to the `cythonize` function in `setup.py` (don't know why these weren't included before. @sprillo?


Lioscro:
Btw, this issue started happening across multiple packages (such as HDBSCAN, sklearn, etc. that use Numpy within Cython) after Numpy 1.22.0 was released on Dec 31, 2021.

mattjones315 (approved):
This looks good to me, thanks so much Joseph for handling this!

PR #163, opened 2022-01-17
remove logging message in error_correct_intbcs

Lioscro:
A lot of logging is spit out in `filter_molecule_table` due to a logging message that prints out every corrected intBC. This PR removes that.


mattjones315 (approved):
Looks good to me! Thanks.

PR #164, opened 2022-01-18
Bam group by speedup

Lioscro:
Using `hits.utilities.group_by` happens to be very slow for some reason, so I implemented our own `group_bam_by_key` function, which achieves approx. 100x speedup.

Branches from `remove-intbc-logging` #163


colganwi:
For me this PR does not speedup `cas.pp.collapse` at all. @Lioscro can you provide a test case or more details about the compute environment? Maybe this is faster with memory constraints?

Lioscro:
It's been a hot minute since I've worked on this, so I'm having a hard time providing any additional context, but I believe the speedup was achieved for very large BAMs.

But given my limited recollection, I'd trust your experience more than whatever I can remember, so I'm fine if you decide this PR is no longer needed.

colganwi:
I tested with a 15GB bam file and see a 2X speed up. MD5 checksums of the outputs are the same. Nice work @Lioscro!

colganwi (approved):
Looks good to me!

PR #168, opened 2022-02-01
Fix incorrect umi count displayed in error_correct_umis

Lioscro:
Note that this is only a visual (logging) bug.

Branched from #163


mattjones315 (approved):
Well, I'm embarrassed that it's taken me this long to get to this. But it looks good! Let's merge it in.

PR #175, opened 2022-02-24
Update name to cassiopeia-lineage

adamgayoso:
Didn't get the name right on the first try -- this is important when uploading to pypi


mattjones315 (approved):

mattjones315 (changes requested):
Actually this is not good to go - looks like this has broken some things around the namespace.

adamgayoso:
@mattjones315 should be good now

mattjones315 (approved):

PR #176, opened 2022-02-28
Update README.md installation

adamgayoso:
This is better than before for new users. Someone should check it really works!


mattjones315 (approved):
I checked this out and it seems to work for me - thanks Adam!

PR #177, opened 2022-03-03
refactored how hybrid log files are created

mattjones315:
This PR fixes an issue identified by a user that had ~150 characters in their data: in this case, this created logfiles for HybridSolvers that were too long for linux systems. This PR sets up a simple iterator for creating new logfiles, and stores the root of the subproblem within the log.


Lioscro (approved):
LGTM!

PR #180, opened 2022-05-18
speed up compute_dissimilarity_map

Lioscro:
Speed up `cas.data.CassiopeiaTree.compute_dissimilarity_map` by only computing dissimilarities between unique states first, and then expanding the dissimilarity map back to all cells.


cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1883,12 +1888,26 @@ def compute_dissimilarity_map(
...
+        dissimilarity_cells = list(dedup_character_matrix.index)
+        j = N
+        for i, dedup_cell in enumerate(dedup_character_matrix.index):
+            for cell in state_to_cells[cell_to_state[dedup_cell]]:
+                if dedup_cell == cell:
+                    continue
+                dissimilarities = full_dissimilarity_map[i]
+                full_dissimilarity_map[j] = dissimilarities
+                full_dissimilarity_map[:, j] = dissimilarities
+                full_dissimilarity_map[j, j] = dissimilarities[i]
```
mattjones315: I don't think we need this line (removing in separate commit).

mattjones315 (approved):
everything looks good to me

PR #182, opened 2022-06-20
Spatial tree plotting

Lioscro:
* `cas.sim.ClonalSpatialDataSimulator` - Assigns spatial coordinates for each leaf while applying a spatial autocorrelation constraint.
* 3D tree visualization with `cas.pl.Tree3D`.


cassiopeia/simulator/ClonalSpatialDataSimulator.py
```diff
@@ -0,0 +1,283 @@
...
+
+        Args:
+            points: Point coordinates
+            k: Number of nearest neighbors
+
+        Returns:
+            Networkx graph
+        """
+        distances = neighbors.kneighbors_graph(points, k, mode='distance')
+        G = nx.from_scipy_sparse_matrix(distances)
```
colganwi, line 118: `from_scipy_sparse_matrix` has been changed to `from_scipy_sparse_array` in Networkx >2.0.0. Needs to be changed before merge

mattjones315: It's actually for Networkx>3.0 I believe. The problem is that Networkx 3.0 and greater is only available for python>=3.8. So, I think it's a natural time to deprecate support for 3.7 (we're behind on this anyways) and just focus on >=3.8.

colganwi:
Replaced hotkeys with depth slider. Works in Jupyter but the update is a bit slow.

Screenshot 2023-08-23 at 11 58 01 AM

colganwi:
Looks good to me! Once the the resolve UMI test failure caused by the version upgrade is dealt with we can merge.

pyproject.toml
```diff
@@ -54,27 +55,36 @@ nbconvert = ">=5.4.0"
...
 parameterized = "*"
 plotly = ">=5.0.0"
+poisson-disc = {version = ">=0.2.1", optional = true}
 pre-commit = {version = ">=2.7.1", optional = true}
 pydata-sphinx-theme = {version = ">=0.4.3", optional = true}
 pysam = ">=0.14.1"
 pyseq-align = ">=1.0.2"
 pytest = {version = ">=4.4", optional = true}
 python = ">=3.7,<4.0"
+pyvista = {version = "=0.41.0", optional = true}
```
colganwi, line 73: I've fixed version for now since there is a bug in 0.41.1. Can change once this is resolved

colganwi (approved):

mattjones315:
Alright this looks good to me too -- 

* I fixed the new errors emerging from updating python and related packages
* I tested the 3D plotter locally and all weird caching errors go away. I presume this has to do with the way Jupyter caches things.

I'm going to make a new issue just reminding us that we need to write a tutorial (or add to the existing plotting tutorial) on how to use the 3D plotter and these new functionalities in general.

PR #190, opened 2023-03-07
Add LBIJungle

sprillo:
Add LBIJungle, which estimates fitness for nodes in the tree.
- I manually added the jungle package under `_jungle`. This adds a lot of files. Note that that Jungle package and the Neher et al. code it wraps is under MIT license. Let me know if we should credit them in some specific way, e.g. in our README.
- See `lbi_jungle_test` for a minimal working example.


cassiopeia/tools/fitness_estimator/_lbi_jungle.py
```diff
@@ -0,0 +1,116 @@
...
+import sys
+import tempfile
+from typing import Optional
+
+import networkx as nx
+
+dir_path = os.path.dirname(os.path.realpath(__file__))
+print(f"dir_path = {dir_path}")
+sys.path.append(os.path.join(dir_path, "_jungle"))
+import jungle as jg
```
sprillo, line 11: This is horrible but I didn't know a better way to do it.

cassiopeia/tools/fitness_estimator/_lbi_jungle.py
```diff
@@ -0,0 +1,116 @@
...
+import numpy as np
+
+from cassiopeia.data import CassiopeiaTree
+
+from ._FitnessEstimator import FitnessEstimator, FitnessEstimatorError
+
+
+def _to_newick(tree: nx.DiGraph, record_branch_lengths: bool = False) -> str:
```
sprillo, line 19: I think this might be a slight variation of something already in our codebase, but tailored to the Jungle format. Or it might be identical, we can check.

cassiopeia/tools/fitness_estimator/_jungle/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+To install jungle, clone this repository to a local directory.
+
+To import jungle, append the jungle directory to your path, then do the import:
+
+```python
+import sys
+sys.path.append("/path/to/jungle")
+import jungle as jg
+```
```
mattjones315, line 41: I'm not sure we should include this part of the README (if we include the README at all)

cassiopeia/tools/fitness_estimator/_jungle/environment.yml
```diff
@@ -0,0 +1,95 @@
+name: jungle
```
mattjones315, line 1: do we need to include Felix's environment.yml here? are there any dependencies here that are needed that we don't have in the regular Cassiopeia requirements?

cassiopeia/tools/fitness_estimator/_lbi_jungle.py
mattjones315: Ha, yeah this isn't pretty but unless we make jungle a module (adding an `__init__.py` etc) then we can't import it directly, right? We can keep this here as is for now.

cassiopeia/tools/fitness_estimator/_lbi_jungle.py
mattjones315: I think this is the same as `cassiopeia.data.utilities.to_newick`

mattjones315 (changes requested):
Thanks again for implementing this @sprillo !

This looks great. Just three small comments regarding if (a) we should keep the README for the internal packages as they are (it might be confusing to users); (b) if we should keep the environment.yml for jungle or not; and (c) the use of the `_to_newick` function in the LBIJungle file (I think it's already implemented in `cassiopeia.data.utilities`)

Happy to merge it in once we get on the same page with these two items.

cassiopeia/tools/fitness_estimator/_lbi_jungle.py
sprillo: It seems like they are slightly different, to account for internal node names. The original version in `cassiopeia.data.utilities.to_newick` has:
```
        if record_branch_lengths and g.in_degree(node) > 0:
            parent = list(g.predecessors(node))[0]
            weight_string = ":" + str(g[parent][node]["length"])
```
whereas the version I wrote here has two more lines:
```
        if record_branch_lengths and g.in_degree(node) > 0:
            parent = list(g.predecessors(node))[0]
            weight_string = ":" + str(g[parent][node]["length"])
            if not is_leaf:
                weight_string = node + weight_string
```
I suggest we just keep these two separate versions. Not worth trying to consolidate them. The one I wrote is a private function in a private module anyway.

cassiopeia/tools/fitness_estimator/_lbi_jungle.py
sprillo: I think I had originally tried with `__init__.py` but it didn't work for some reason. So I just used this. Agree it's fine to keep as is.

cassiopeia/tools/fitness_estimator/_jungle/README.md
sprillo: I'll just ahead and remove this stuff. I literally had just added the whole repo.

sprillo:
@mattjones315 I addressed your points. I removed the README and yml file from `_jungle`. I left `_to_newick` since it is slightly different.

Additionally, I added a check that is needed due to the underlying implementation we are wrapping.... leaf names cannot start with an underscore. Check the new test I wrote. (Of course, we could map leaf names to something new without underscore to run LBI jungle, and then unmap, but I'm too lazy to do that now and test it). I don't think anyone will ever face this issue anyway.

sprillo:
(Sorry had forgotten to push.)

cassiopeia/tools/fitness_estimator/_lbi_jungle.py
mattjones315: Got it, I agree it's fine to keep.

cassiopeia/tools/fitness_estimator/_jungle/README.md
mattjones315: Great, thanks!

mattjones315 (approved):
Looks good to me, let's merge it in. Thanks Sebastian!

PR #192, opened 2023-03-14
Implement vectorized log-likelihood computation to speed up cvxpy.

sprillo:
As title.

With the new implementation, branch length estimation should be ~10-20x faster.

I have extended `IIDExponentialMLE`'s API with a private argument that toggles between the vectorized and non-vectorized implementations, to be able to profile the difference.

As I have observed in the past, the ECOS solver benefits the most from vectorization (SCS instead performs the same...). Because of this, I have decided to make the ECOS solver (with vectorization) the default one. In this sense, this PR is non-backwards compatible - since the default solver used to be SCS - but it's fine since out method has not been published yet. Unfortunately, the ECOS solver is not as stable as SCS and can sometimes fails (say 5% of the time). Because of this, I have extended the API of the branch length estimator to include a `backup_solver`. If the main `solver` fails, we retry with the backup solver. With the new API's default arguments, the main solver is ECOS, and if it fails, the slower but reliable SCS solver will be used.

NOTE: switching to the ECOS solver on its own speeds up the branch length estimator by ~7x, the remaining ~2-3x being provided by vectorization. The reason I did not use ECOS originally is because SCS is extremely robust; however, we want something faster for dealing with site rate variation now, so the new scheme of ECOS + fallback to SCS should be awesome.

Test plan:
- All old tests pass.
- The following code snippet profiles branch length estimation on a tree with 512 leaves, using either SCS or ECOS, and a vectorized or non-vectorized implementation (for a total of 4 combinations):
```
import time

from cassiopeia.simulator import CompleteBinarySimulator, Cas9LineageTracingDataSimulator
from cassiopeia.tools.branch_length_estimator import IIDExponentialMLE

tree = CompleteBinarySimulator(num_cells=512).simulate_tree()
Cas9LineageTracingDataSimulator(mutation_rate=1.0, random_seed=0).overlay_data(tree)
for solver in ["SCS", "ECOS"]:
    for _use_vectorized_implementation in [False, True]:
        st = time.time()
        IIDExponentialMLE(
            minimum_branch_length=0.01,
            pseudo_mutations_per_edge=0.1,
            pseudo_non_mutations_per_edge=0.1,
            solver=solver,
            _use_vectorized_implementation=_use_vectorized_implementation,
        ).estimate_branch_lengths(tree)
        et = time.time()
        msg_str= f"Time using solver {solver} and _use_vectorized_implementation={_use_vectorized_implementation}: {et - st}"
        print(msg_str)
```
Output:
```
Time using solver SCS and _use_vectorized_implementation=False: 52.351640939712524
Time using solver SCS and _use_vectorized_implementation=True: 46.90113306045532
Time using solver ECOS and _use_vectorized_implementation=False: 6.936963081359863
Time using solver ECOS and _use_vectorized_implementation=True: 2.934386968612671
```

### DIFF
--- a/cassiopeia/tools/branch_length_estimator/_iid_exponential_mle.py
+++ b/cassiopeia/tools/branch_length_estimator/_iid_exponential_mle.py
@@ -64,6 +64,14 @@ class IIDExponentialMLE(BranchLengthEstimator):
             None is the default value for this argument.
         solver: Convex optimization solver to use. Can be "SCS", "ECOS", or
             "MOSEK". Note that "MOSEK" solver should be installed separately.
+            We recommend "ECOS" (which is the default).
+        backup_solver: In case the main solver fails, this backup solver will
+            be tried instead. Useful for applying a faster but less
+            sophisticated solver first, and only falling back to a very
+            reliable but slower solver if needed. We recommend "SCS" (which is
+            the default). (If `backup_solver=None` is provided, no retry will be
+            attempted and an error will be raised immediately if the main solver
+            fails.)
         _use_vectorized_implementation: Toggles between vectorized and
             non-vectorized implementations. Only used for profiling purposes.
         verbose: Verbosity level.
@@ -94,6 +102,7 @@ class IIDExponentialMLE(BranchLengthEstimator):
         relative_mutation_rates: Optional[List[float]] = None,
         verbose: bool = False,
         solver: str = "ECOS",
+        backup_solver: Optional[str] = "SCS",
         _use_vectorized_implementation: bool = True,
     ):
         allowed_solvers = ["ECOS", "SCS", "MOSEK"]
@@ -109,10 +118,12 @@ class IIDExponentialMLE(BranchLengthEstimator):
         self._relative_mutation_rates = relative_mutation_rates
         self._verbose = verbose
         self._solver = solver
+        self._backup_solver = backup_solver
         self._use_vectorized_implementation = _use_vectorized_implementation
         self._mutation_rate = None
         self._penalized_log_likelihood = None
         self._log_likelihood = None
+        self._backup_solver_was_needed = None
 
     def estimate_branch_lengths(self, tree: CassiopeiaTree) -> None:
         r"""
@@ -134,6 +145,7 @@ class IIDExponentialMLE(BranchLengthEstimator):
         relative_leaf_depth = self._relative_leaf_depth
         relative_mutation_rates = self._relative_mutation_rates
         solver = self._solver
+        backup_solver = self._backup_solver
         _use_vectorized_implementation = self._use_vectorized_implementation
         verbose = self._verbose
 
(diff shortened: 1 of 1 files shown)


cassiopeia/tools/branch_length_estimator/_iid_exponential_mle.py
```diff
@@ -292,8 +365,17 @@ def estimate_branch_lengths(self, tree: CassiopeiaTree) -> None:
...
+        except cp.SolverError as err:  # pragma: no cover
+            # We try the backup_solver
+            try:
+                if backup_solver is None:
+                    # We don't retry; just raise the original error
+                    raise err
+                prob.solve(solver=backup_solver, verbose=verbose)
+                self._backup_solver_was_needed = True
+            except cp.SolverError:  # pragma: no cover
+                raise IIDExponentialMLEError("Third-party solver failed")
```
scimind1729, line 378: Maybe specifically mention that the backup failed? Or if you think this is sufficiently covered by this coupled with _backup_solver_was_needed then it's fine.

scimind1729, line 378: Maybe specifically mention that the backup failed? Or if you think this is sufficiently covered by this coupled with _backup_solver_was_needed then it's fine.

scimind1729 (changes requested):
LGTM. Might be worth considering comparing the output of the vectorized ECOS and unvectorized SCS solvers on some larger trees to see if they agree, even if we can't solve them by hand.

sprillo:
> LGTM. Might be worth considering comparing the output of the vectorized ECOS and unvectorized SCS solvers on some larger trees to see if they agree, even if we can't solve them by hand.

This is a good point. In the notebook above, the trees all look identical. For the record, this code:
```
import time
import cassiopeia as cas

from cassiopeia.simulator import CompleteBinarySimulator, Cas9LineageTracingDataSimulator
from cassiopeia.tools.branch_length_estimator import IIDExponentialMLE

tree = CompleteBinarySimulator(num_cells=512).simulate_tree()
Cas9LineageTracingDataSimulator(mutation_rate=1.0, random_seed=0).overlay_data(tree)
for solver in ["SCS", "ECOS"]:
    for _use_vectorized_implementation in [False, True]:
        st = time.time()
        IIDExponentialMLE(
            minimum_branch_length=0.01,
            pseudo_mutations_per_edge=0.1,
            pseudo_non_mutations_per_edge=0.1,
            solver=solver,
            _use_vectorized_implementation=_use_vectorized_implementation,
        ).estimate_branch_lengths(tree)
        et = time.time()
        msg_str= f"Time using solver {solver} and _use_vectorized_implementation={_use_vectorized_implementation}: {et - st}"
        print(msg_str)
        cas.pl.plot_matplotlib(
            tree,
            add_root=False,
            depth_key="time",
            extend_branches=False,
        )
```

We can also easily verify that performance is the same in our benchmarking, by trying out the two versions.

scimind1729 (approved):

PR #205, opened 2023-07-27
Fast UPGMA and NJ with CCPhylo

colganwi:
The C implementations of UPGMA and NJ in [CCPhylo](https://bitbucket.org/genomicepidemiology/ccphylo/src/master/) are significantly faster than the current Cassiopeia implementations of these algorithms. This update adds a `CCPhyloSolver` subclass of `DistanceSolver` which writes the dissimilarity matrix to a temp file and then calls `ccphlyo tree` to solve the tree with the specified method. `NeighborJoiningSolver` and `UPGMASolver` are now subclasses of `CCPhyloSolver` so when fast is set to true they use `CCPhyloSolver._fast_solve()` instead of `DistanceSolver.solve()`.  CCPhylo also includes two new algorithms Dynamic and Heuristic NJ which are significantly faster than standard NJ. `DynamicNeighborJoiningSolver` and `HeuristicNeighborJoiningSolver` are subclasses of `NeighborJoiningSolver` which implement these algorithms. In most cases DNJ should be used since it is guaranteed to generate an exact NJ tree.


colganwi:
Benchmarking results

benchmarking

cassiopeia/solver/CCPhyloSolver.py
```diff
@@ -0,0 +1,223 @@
...
+    solver_utilities,
+)
+
+
+class CCPhyloSolver(DistanceSolver.DistanceSolver):
+    """
+    Distance based solver class. This solver serves as a wrapper around 
+    CCPhylo algorithms.
+
+    Args:
```
mattjones315, line 35: We'll have to add descriptions for the two new parameters here - `fast` and `method`.  We should also be clear about which methods are accepted here, I think we can use the `Literal` type here. (https://docs.python.org/3/library/typing.html#typing.Literal)

cassiopeia/solver/CCPhyloSolver.py
```diff
@@ -0,0 +1,223 @@
...
+    ):
+
+        super().__init__(prior_transformation)
+
+        self.dissimilarity_function = dissimilarity_function
+        self.add_root = add_root
+        self.fast = fast
+        self.method = method
+        
+    def _setup_ccphylo(self) -> None:
```
mattjones315, line 79: nit: we have used a line limit of 80 characters in the past. Would you be able to reformat the strings below to abide by this style guide? If you use VS Code, I've normally set the line limit in VS Code which helps.

cassiopeia/solver/CCPhyloSolver.py
```diff
@@ -0,0 +1,223 @@
...
+        prior_transformation: str = "negative_log",
+        fast = True, 
+        method: str = "nj",
+    ):
+
+        super().__init__(prior_transformation)
+
+        self.dissimilarity_function = dissimilarity_function
+        self.add_root = add_root
+        self.fast = fast
```
mattjones315, line 76: I suggest we call `_setup_ccphylo` in the constructor if `fast=True`.  The reason being is that I want to alert the user as soon as possible if there's an error in the configuration, and that I don't imagine the configuration changing between successive runs of the solver.

cassiopeia/solver/CCPhyloSolver.py
```diff
@@ -0,0 +1,223 @@
...
+            logfile: File location to log output. Not currently used.
+        """
+
+        if self.fast:
+            self._fast_solve(cassiopeia_tree, layer, collapse_mutationless_edges, logfile)
+        else:
+            if self.__class__ == CCPhyloSolver:
+                raise NotImplementedError(
+                    "CCPhyloSolver does not implement solve. Please set fast to True" 
+                    "to use the fast_solve method. Or use a subclass of CCPhyloSolver."
```
mattjones315, line 221: This is weird to me, especially since every method should implement `solve` without throwing an error. I think that this stems from the new structure of this PR (i.e., NeighborJoiningSolver inheriting from CCPhyloSolver). I wonder if we should consider getting rid of the CCPhyloSolver class and moving the `_fast_solve` logic to the `DistanceSolver` class. Then, if `fast=True` in the `DistanceSolver`, we'll divert the process to the `_fast_solver`; else we'll stay with the normal `solve` method.

cassiopeia/solver/CCPhyloSolver.py
```diff
@@ -0,0 +1,223 @@
...
+            # save dissimilarity map as phylip file
+            dis_path = os.path.join(temp_dir, "dist.phylip")
+            tree_path = os.path.join(temp_dir, "tree.nwk")
+            solver_utilities.save_dissimilarity_as_phylip(dissimilarity_map, dis_path)
+
+            # run ccphylo
+            command = f"{self.ccphylo_path} tree -i {dis_path} -o {tree_path} -m {self.method}"
+            process = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            stdout, stderr = process.communicate()
+            T = ete3.Tree(tree_path, format=1)
```
mattjones315, line 142: Is there a reason here why we can't read in the newick file with the Cassiopeia API? We have a few functions implemented under `tree.data.utilities` for reading in newick trees (and you can also pass the newick string to the `CassiopeiaTree` constructor)

cassiopeia/solver/HeuristicNeighborJoiningSolver.py
```diff
@@ -0,0 +1,67 @@
...
+    Attributes:
+        dissimilarity_function: Function used to compute dissimilarity between
+            samples.
+        add_root: Whether or not to add an implicit root the tree.
+        prior_transformation: Function to use when transforming priors into
+            weights.
+
+    """
+
+    def __init__(
```
mattjones315, line 49: need to make sure we add `fast` to the set of arguments here, right? Same with other subclasses of `CCPhyloSolver`.

cassiopeia/solver/solver_utilities.py
```diff
@@ -122,3 +123,24 @@ def convert_sample_names_to_indices(
     name_to_index = dict(zip(names, range(len(names))))
 
     return list(map(lambda x: name_to_index[x], samples))
+
+def save_dissimilarity_as_phylip(
```
mattjones315, line 127: we should add a test for this, it can be under `test/solver_tests/ccphylo_solver_test.py` if that's easiest.

mattjones315 (changes requested):
Thanks for this fantastic PR @colganwi! It's looking really good, and I appreciate you learning our code structure and style guide so quickly as to make contributions.

Most of the code looks great, minus some nitpicky issues here and there around documentation and style (sorry, I know it's annoying).

One major issue that I'd like to get hammered out, and perhaps you can just convince me of it, is that I'm not sure we need to have a separate `CCPhyloSolver` class. I feel like there are some quirks about it that make it feel half-completed, such as the fact that you run the risk of throwing an error with `solve`. I think there are two potential solutions, if you are open to them:

1. I think we can keep `solve` as an abstractmethod (marked with the `@abc.abstractmethod` decorator) in `CCPhyloSolver`. We can keep the `_fast_solve` method implemented. Then, in any of the subclasses (e.g., `NeighborJoiningSolver`), we can override `solve` as to take into account the `fast` variable. In other words, if `fast=True`, we call `super()._fast_solve`; else we call `super().super().solve()`.
2. Second, I think we can get rid of `CCPhyloSolver` altogether and move the logic to `DistanceSolver`. In fact, that would make the first suggestion more coherent as well. We'd just have to move `_fast_solve` to `DistanceSolver`. Then, all subclasses would be invoked with `solve`, and we can encode the decision between `_fast_solve` and the traditional `solve` logic within `DistanceSolver`.

Do either of these suggestions make sense? Happy to discuss more in depth, and happy to be convinced that we need `CCPhyloSolver` if that's really the case.

mattjones315:
Oh and one more comment -- we should make sure to merge in the latest changes to the `master` branch, as the pandas typing/ILP error appears to still be a problem in running the tests.

sprillo (changes requested):
Thanks @colganwi , this is awesome, I've always wanted a faster NJ implementation!

All the pieces seem to be there, but I want to leave some thoughts on the current design.

I think using inheritance is not the ideal way to implement the "fast" functionality in the distance solvers. Using inheritance (with NeighborJoiningSolver inheriting from CCPhyloSolver) couples NeighborJoiningSolver to the CCPhyloSolver, and in particular to its "fast" implementation. Here's an example of why this is problematic: Suppose that in the future another even faster implementation of NJ comes up, and we want to expose it, but we still want to support the current CCPhyloSolver implementation. What do we do?

I think a superior way to implement the new "fast" functionality is to use composition over inheritance. This requires minimal changes to the current NeighborJoiningSolver class (and _none_ to the DistanceSolver class). The pseudocode would look like this:

```
class NeighborJoiningSolver(DistanceSolver.DistanceSolver):
    def __init__(
        ...
        implementation: str = "ccphylo"
    ):
        self._implementation = implementation
        ...

    def solve(self, ...):
        if self._implementation == "ccphylo":
            solver = CCPhyloSolver(...)  # (Note: Could be done in __init__ instead to raise errors earlier.)
            solver.solve(tree)
        else:
            super().solve(...)
```

This way, NeighborJoiningSolver essentially wraps the CCPhyloSolver, instead of inheriting from it. Also, CCPhyloSolver would implement "solve" directly and not have a "_fast_solve" method. With this approach, if other faster implementations of NJ come up in the future, they can just be registered in the NeighborJoiningSolver's solve method, e.g.:

    def solve(self, ...):
        if self._implementation == "ccphylo":
            solver = CCPhyloSolver(...)
            solver.solve(tree)
        elif self._implementation == "even_faster_implementation":
            solver = EvenFasterNJSolver(...)
            solver.solve(tree)
        else:
            super().solve(...)

I think it's fine to have a concrete CCPhyloSolver class. In fact, one could argue there is no need for modifying NeighborJoiningSolver at all, since one could just instantiate and use the CCPhyloSolver instead. However, as I see it, NeighborJoiningSolver is a "centralized" place to create NJ solvers with slightly different implementations. In  this sense, one could even argue that the DynamicNeighborJoiningSolver and HeuristicNeighborJoiningSolver classes are unnecessary and could be realized through the NeighborJoiningSolver directly, e.g.:

    def solve(self, ...):
        if self._implementation == "ccphylo_nj":
            solver = CCPhyloSolver(...)
            solver.solve(tree)
        elif self._implementation == "ccphylo_hnj":
            solver = CCPhyloSolver(...)  # Use HNJ
            solver.solve(tree)
        elif self._implementation == "ccphylo_dnj":
            solver = CCPhyloSolver(...)  # Use DNJ
            solver.solve(tree)
        else:
            super().solve(...)

Regardless, the main point is that I am not sure going down the inheritance path is ideal because it creates deep, coupled hierarchies and code that is harder to read. @mattjones315 what do you think?

Some resources: 
Composition over inheritance: https://youtu.be/hxGOiiR9ZKg
Strategy pattern: https://en.wikipedia.org/wiki/Strategy_pattern

cassiopeia/solver/CCPhyloSolver.py
colganwi: Good point! I've fixed this.

cassiopeia/solver/CCPhyloSolver.py
colganwi: I like this solution. The CCPhyloSolver class is removed by c0c5ca926624067026e26d5483e49a18b3bc3f80

cassiopeia/solver/CCPhyloSolver.py
colganwi: I want to use a few ete3 methods so I'm sticking with the ete3 reader, but I've switched to using `tree.data.utilities.ete3_to_networkx` to convert the ete3 tree to the nx tree

cassiopeia/solver/HeuristicNeighborJoiningSolver.py
colganwi: Do we want to have a fast parameter here? There is no slow implementation of the HNJ so we would need to throw an error if fast = False

colganwi:
Unfortunately, I implemented @mattjones315's suggestion of removing the CCPhyloSolver class (c0c5ca926624067026e26d5483e49a18b3bc3f80) before seeing @sprillo's comments. This commit addresses the issue of a future "even_faster_implementation" faster implementation by changing `CCPhyloSolver._fast_solve` to `DistanceSolver._ccphylo_solve`. The fast solver is then selected in `DistanceSolver.__init__`  which would allow other fast solver implementations. 

But now that I've read @sprillo's comment I think his solution is cleaner since it doesn't require modifications to the DistanceSolver class and would allow for CCPhylo methods such as K-means Closest First to be run without creating a dedicated `KMeansClosestFirstSolver` subclass of the `DistanceSolver`. However, one issue with this approach is rooting the tree. Currently, the CCPhyloSolver versions of UPGMA and NJ use different rooting strategies to ensure that the results are the same as the existing implementations. It would be redundant to also put these rooting strategies into the `CCPhyloSolver` class. Of course, the obvious solution to this is to make the rooting functions compositional but this will require some work/reorganization. Do we want to make this change to how the solvers are organized?

sprillo (approved):
@colganwi given the current design used for solvers in Cassiopeia I think the solution you have currently implemented (diverting to a "_fast_solve" variant in `DistanceSolver`) is okay.

I think refactoring (or, completely redesigning the solver package) isn't worth the effort right now, but I agree that the current design using inheritance has a few issues. Generally, the solver classes take on too many responsibilities / make too many assumptions. For example, the abstract base class `CassiopeiaSolver` is not only responsible for estimating a tree topology but also for collapsing mutationless edges. Partly as a results you see a lot of code duplication in the concrete subclasses, e.g. in [DistanceSolver](https://github.com/YosefLab/Cassiopeia/blob/master/cassiopeia/solver/DistanceSolver.py#L194-L198) , [GreedySolver](https://github.com/YosefLab/Cassiopeia/blob/master/cassiopeia/solver/GreedySolver.py#L180-L184) , [ILPSolver](https://github.com/YosefLab/Cassiopeia/blob/master/cassiopeia/solver/ILPSolver.py#L263-L267). (This code could be moved to the parent class somehow, but that's not the point, the point is solvers probably should not have to bother about collapsing mutationless edges at all).

In the particular case of the `DistanceSolver` class, subclasses are expected to conform to a specific algorithm structure, providing hooks for `find_cherry`, `update_dissimilarity_map`, `setup_root_finder`, `root_tree` that are used in `DistanceSolver`'s `solve` method. This avoids duplicating the `DistanceSolver`'s `solve` logic in the subclasses, but it limits the kind of `DistanceSolver`'s that are expected. In this sense, the `CCPhyloSolver` class is not a `DistanceSolver`, since a `CCPhyloSolver` class wrapping the C implementation actually reuses _none_ of `DistanceSolver`'s "solve" method and implements none of the hooks. It is simply a `CassiopeiaSolver`. For this reason, in a PR I recently opened ( https://github.com/YosefLab/Cassiopeia/pull/211 - feel free to leave comments ) where I implement what is quite intuitively a distance-based method, I don't inherit from `DistanceSolver` at all.

To summarize, a distance solver wrapping a C++ implementation shouldn't really be thought of as a `DistanceSolver` given what the current expectations of the `DistanceSolver` class are. Inheriting from `DistanceSolver` and then overriding `solve` to divert to a `_fast_solve` (as in the original PR) is misleading from this perspective. The current solution, where `DistanceSolver` allows for a `_fast_solve` seems more readable to me. However, note that `DistanceSolver` is now basically doing two different things: (1) providing a skeleton `solve` as originally which you can hook to, and (2) knowing how to run ccphylo or other programs. So it's basically responsible for two completely unrelated things. I don't mind it too much as it is, but you can see how it could hamper readability and make things a bit confusing for a newcomer.

I'll unblock the PR on my end now that we are aware of the pros and cons of the current design. I'll let Matt have the final word on what shape this should have.

cassiopeia/solver/DistanceSolver.py
```diff
@@ -11,11 +11,17 @@
 import networkx as nx
 import numpy as np
 import pandas as pd
+import configparser
+import subprocess
+import os
+import tempfile
+import ete3
```
mattjones315, line 19: nit: these imports should follow our Cassiopeia style - namely, `os` and `sys` should be imported first; then system-level imports; then packages installed manually. All should be alphabetical.

For example,

```
import os

import subprocess
import tempfile

import configparser
import ete3
import networkx as nx
import numpy as np
import pandas as pd
```

cassiopeia/solver/DistanceSolver.py
```diff
@@ -135,6 +157,12 @@ def solve(
                 removes artifacts caused by arbitrarily resolving polytomies.
             logfile: File location to log output. Not currently used.
         """
+
+        # Use fast solver if selected
+        if self.fast:
+            self.fast_solve(cassiopeia_tree,layer,collapse_mutationless_edges,logfile)
```
mattjones315, line 163: nit: line should be less than 80 characters long

cassiopeia/solver/DistanceSolver.py
```diff
@@ -197,6 +225,112 @@ def solve(
...
+        """
+
+        dissimilarity_map = self.get_dissimilarity_map(cassiopeia_tree, layer)
+
+        with tempfile.TemporaryDirectory() as temp_dir:
+
+            # save dissimilarity map as phylip file
+            dis_path = os.path.join(temp_dir, "dist.phylip")
+            tree_path = os.path.join(temp_dir, "tree.nwk")
+            solver_utilities.save_dissimilarity_as_phylip(dissimilarity_map, dis_path)
```
mattjones315, line 258: nit: line needs to be less than 80 characters long.

cassiopeia/solver/DistanceSolver.py
```diff
@@ -11,11 +11,17 @@
...
+import tempfile
+import ete3
 
 from cassiopeia.data import CassiopeiaTree
 from cassiopeia.mixins import DistanceSolverError
-from cassiopeia.solver import CassiopeiaSolver, solver_utilities
-
+from cassiopeia.solver import (CassiopeiaSolver, 
+    solver_utilities)
+from cassiopeia.data.utilities import ete3_to_networkx
```
mattjones315, line 24: Better to import `from cassiopeia.data import utilities as data_utilities` and then below call `data_utilities.ete3_to_networkx`. Just improves readability for users not familiar with the codebase.

cassiopeia/solver/HeuristicNeighborJoiningSolver.py
mattjones315: You're right, no need -- the current implementation makes sense.

cassiopeia/solver/NeighborJoiningSolver.py
```diff
@@ -1,5 +1,5 @@
 """
-This file stores a subclass of DistanceSolver, NeighborJoining. The
+This file stores a subclass of CCPhyloSolver, NeighborJoiningSolver. The
```
mattjones315, line 2: This is no longer accurate, right? Here `NeighborJoiningSolver` inherits from `DistanceSolver`.

cassiopeia/solver/solver_utilities.py
mattjones315: Still need a test for this function.

mattjones315 (changes requested):
Hi @colganwi and @sprillo --

First, thank you both for a very thoughtful discussion on this PR. It has certainly sparked a lot of good ideas on how to improve functionality and readability, and it is evident to me that we have outgrown our initial design/vision from a few years ago when we executed the first refactor (trust me, this is a much improved version of the codebase back in 2019...).

I like several aspects of Sebastian's proposed refactor towards a compositional structure in general. I certainly see how it would benefit the current implementation of our `DistanceSolver`'s but I also see that these themes can be extended to apply to several of our solvers. In fact, it sounds like given our recent work implementing several new algorithms that the `solver` module is due for a refactor in general. For this reason, I don't want to put the burden on this PR; it deserves a larger effort and more conversation.

For the time being, this PR implements well the suggestions I had earlier this week and I think it will be ready to merge in once my nit-picky comments on formatting and additional tests are addressed. I also want to commend William for making a very solid contribution to the Cassiopeia codebase!

I'm going to make an issue linking this PR around reformatting the `solver` module.

colganwi:
@sprillo I completely agree that the current architecture of the solvers, particularly the `DistanceSolver`, is not ideal. The main reason I built the `CCPhyloSolver` on top the `DistanceSolver` was to take advantage of the logic for rooting trees and calculating dissimilarity maps. Let's plan to address this when we refactor the solvers. We could either create a more generic distance `DistanceSolver`, or with a sufficiently compositional framework, write a `CCPhyloSolver` that does not inherent from `DistanceSolver`. When @mattjones315 opens the issue we can have a more in depth conversation there.  

@mattjones315 thanks for the additional comments. I'm going to add two more commits, one addressing the comments and adding tests, and one with a few new features including multithreading and subclasses for the other solvers implemented by CCPhylo. Once those commits are reviewed we can merge.

mattjones315:
Hi @colganwi, sounds like a great plan. Looking forward to those commits.

I've now opened an issue referencing our thoughts on refactoring the `DistanceSolver` and the `solver` module in general which you can find in #214. William, once you accept the invitation to join the `Cassiopeia` repo as a collaborator I'll add you to the list of assignees (currently just me and Sebastian).

cassiopeia/solver/CCPhyloSolver.py
colganwi: I'm going to stick with str for this with subsequent value checking so I can raise a more informative error.

colganwi:
Should be good to go. Since we are planning to refactor I've minimized changes to the solver class structure. The DNJ and HNJ classes have been removed in favor of an implementation parameter in the `NeighborJoiningSolver` constructor and the `DistanceSolver` constructor is now unchanged. One advantage of this approach is that CCPhylo DNJ is now the default fast implementation of NeighborJoining.

I've also decided to table adding multithreading and classes for other CCPhylo solvers for now to minimize work when we refactor. If we bring back the CCPhyloSolver class these options can be built into it.

mattjones315 (approved):
This looks great! Fantastic work putting together this PR, and I'm glad this was also an opportunity to think about how we can refactor the logic of the `solver` module altogether. I'm just going to make one small commit re-ordering imports to abide by my nitpicky ordering preferences and then we can merge this in.

PR #215, opened 2023-08-23
Spatial leaf subsampler

colganwi:
New `SpatialLeafSubsampler` class uniformly samples leaves within a region of interest defined by a bounding box or mask using the 'spatial' attribute stored in the tree. The number of leaves to sample can be set explicitly or as a ratio of the number of cells in the region. This class is useful for sampling adjacent slices from a 3D spatial simulation.


cassiopeia/simulator/SpatialLeafSubsampler.py
```diff
@@ -0,0 +1,211 @@
...
+                else:
+                    leaf_keep.append(leaf)
+            
+        if space is not None:
+            # Check the dimensions of coordinates and space match
+            if len(coordinates) != len(space.shape):
+                raise LeafSubsamplerError(
+                    f"Dimensions of coordinates ({len(coordinates)}) and "
+                    f"space ({len(space.shape)}) do not match."
+                )
```
mattjones315, line 147: Do you want to place any checks on the size of the masking array?  

I really like this addition to the sampler and would like to propose an additional behavior based on this performance -- how about we allow spots to be merged together depending on the size of the masking array? In this way, we can have a lot of control over simulating datasets of different resolution by maintaining the same overall shape but increasing / decreasing the size of the array.

Of course there might be weird behavior if for example a particular leaf's coordinates overlaps two or more adjacent spots, but we're not simulating UMIs here so we can just let the leaf appear in these spots it overlaps (as long it overlaps above a certain threshold, perhaps).

cassiopeia/simulator/SpatialLeafSubsampler.py
```diff
@@ -0,0 +1,211 @@
...
+import numpy as np
+from typing import Optional, List
+
+from cassiopeia.data import CassiopeiaTree
+from cassiopeia.simulator.LeafSubsampler import (
+    LeafSubsampler,
+    LeafSubsamplerError,
+)
+from cassiopeia.mixins import (
+    CassiopeiaTreeError
```
mattjones315, line 19: nit: alphabetize imports (place `cassiopeia.mixins` above `cassiopeia.simulator`)

cassiopeia/simulator/SpatialLeafSubsampler.py
```diff
@@ -0,0 +1,211 @@
...
+A subclass of LeafSubsampler, the SpatialLeafSubsampler. 
+
+Uniformly samples leaves within a region of interest of a CassiopeiaTree
+with spatial information and produces a new CassiopeiaTree containing the
+sampled leaves. 
+"""
+
+import copy
+import numpy as np
+from typing import Optional, List
```
mattjones315, line 11: nit: imports should be:

```
from typing import List, Optional

import copy
import numpy as np
```

cassiopeia/simulator/SpatialLeafSubsampler.py
```diff
@@ -0,0 +1,211 @@
...
+        attribute_key: str = "spatial"
+    ):
+        """
+        Subsets the leaves of a CassiopeiaTree to those within a region of
+        interest. The region of interest is defined by a bounding box, which
+        is a list of tuples of the form (min, max) for each dimension. If 
+        'number_of_leaves' is provided, 'number_of_leaves' of the leaves are
+        sampled uniformly from within the bounding box.
+
+        Args:
```
mattjones315, line 38: Nit: add `Raises` section for completeness.

cassiopeia/simulator/SpatialLeafSubsampler.py
```diff
@@ -0,0 +1,211 @@
...
+        self.__bounding_box = bounding_box
+        self.__space = space
+        self.__ratio = ratio
+        self.__number_of_leaves = number_of_leaves
+        self.__attribute_key = attribute_key
+
+    def subsample_leaves(
+        self, tree: CassiopeiaTree, keep_singular_root_edge: bool = True
+    ) -> CassiopeiaTree:
+        """
```
mattjones315, line 83: Add a short one-liner description that will help with the documentation.

mattjones315 (changes requested):
Looking great! I left a few small nit-picky comments that should be fixed up before merging in. I also left a more substantial question regarding the use of the `space` argument & behavior. Looking forward to your thoughts.

cassiopeia/simulator/SpatialLeafSubsampler.py
colganwi: Yes, I've added a warning for when the max coordinate is much smaller than the shape of the space

I've added spatial downsampling. Let me know what you think. The current implementation does not rescale the coordinated based on the shape of the space parameter so the user will have to rescale the spatial attribute to test many different resolutions. I did this because its more intuitive and I don't want to distort the shape but happy to automatically rescale or add a rescaling flag if you want this functionality.

I don't see how leaf coordinates overlapping adjacent spots is an issue since each leaf is a represented by a point and the spots are non-overlapping. Am I missing something?

mattjones315: Great!

I think it makes sense to expect the user to rescale the space coordinates to match the shape of the sampler.

As for overlapping spots/leaves, I think it's an issue right now given the way we are modeling leaves. In the future, we can think about specifying a spherical shape for each leaf to model how a cell would really look in a spatial transcriptomics experiment. When we do, we will have to think about how cells can participate in multiple spots.

test/simulator_tests/spatial_leaf_subsampler_test.py
```diff
@@ -0,0 +1,249 @@
...
+import numpy as np
+import pandas as pd
+
+from cassiopeia.data.CassiopeiaTree import CassiopeiaTree
+from cassiopeia.simulator.LeafSubsampler import (
+    LeafSubsamplerError, 
+    LeafSubsamplerWarning
+)
+from cassiopeia.simulator.SpatialLeafSubsampler import SpatialLeafSubsampler
+
```
mattjones315, line 13: nit: need to have two lines between imports and code body. I can make this small change for you.

mattjones315 (approved):
Looks great! There's just one small nit picky formatting issue in the test; I'll make this change and merge it in to the codebase.

PR #216, opened 2023-08-23
Generalization of Cas9LineageTracingDataSimulator

colganwi:
Makes `Cas9LineageTracingDataSimulator` more general by adding `self.mutation_priors_per_character` which allows for different characters to have different mutation distributions. The API is the same, except `mutation_priors` can now be a list of length `size_of_cassette` or `number_of_cassettes * size_of_cassette`. For consistency, `mutation_rate` now also excepts lists of `size_of_cassette` or `number_of_cassettes * size_of_cassette` instead of just `number_of_cassettes * size_of_cassette`.


cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -73,15 +73,17 @@ class Cas9LineageTracingDataSimulator(LineageTracingDataSimulator):
         size_of_cassette: Number of editable target sites per cassette
         mutation_rate: Exponential parameter for the Cas9 cutting rate. The user
             can either pass in a float, in which every site will mutate at
-            the same rate, or a list with an entry for each cut site indicating
-            potentially a different mutation rate for every site.
+            the same rate, or a list a list of floats of length
```
mattjones315, line 76: typo: "or a list a list of floats" -> "or a list of floats of length..."

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -73,15 +73,17 @@ class Cas9LineageTracingDataSimulator(LineageTracingDataSimulator):
         size_of_cassette: Number of editable target sites per cassette
         mutation_rate: Exponential parameter for the Cas9 cutting rate. The user
             can either pass in a float, in which every site will mutate at
-            the same rate, or a list with an entry for each cut site indicating
-            potentially a different mutation rate for every site.
+            the same rate, or a list a list of floats of length 
+            `size_of_cassette` or `number_of_cassettes * size_of_cassette`.
```
mattjones315, line 77: Please specify the expected functionality of each variant of these arguments -- i.e., we assume that if the `float` is passed in then all sites have the same mutation rate; if just a list of `size_of_cassette` then rates differ across a cassette but each cassette is the same etc.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -73,15 +73,17 @@ class Cas9LineageTracingDataSimulator(LineageTracingDataSimulator):
...
         state_generating_distribution: Distribution from which to simulate state
             likelihoods. This is only used if mutation priors are not
             specified to the simulator.
         number_of_states: Number of states to simulate
-        mutation_priors: A mapping from state to probability that a user
-            can specify. If this argument is not None, states will not be
-            pulled from the state distribution.
+        mutation_priors: An optional dictionary mapping states to their prior 
+            probabilities. Can also be a list of dictionaries of length 
+            `size_of_cassette` or `number_of_cassettes * size_of_cassette`.
```
mattjones315, line 84: same request as above - make it more clear what the expected behavior is for each of these parameter styles.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -132,42 +134,66 @@ def __init__(
...
             ] * number_of_characters
         else:
-            if len(mutation_rate) != (
-                self.number_of_cassettes * self.size_of_cassette
-            ):
+            if len(mutation_rate) == number_of_characters:
+                self.mutation_rate_per_character = mutation_rate
+            elif len(mutation_rate) == self.size_of_cassette:
+                self.mutation_rate_per_character = (
+                    mutation_rate * self.number_of_cassettes
```
mattjones315, line 152: you should mke sure that `mutation_rate` is actually a List. If it's a numpy array, then this line will just multiply the mutation rate by the number of cassettes.

test/simulator_tests/cas9_lineage_tracing_simulator_test.py
```diff
@@ -151,15 +155,15 @@ def test_basic_setup(self):
...
+            self.assertAlmostEqual(
+                1.0,
+                np.sum(
+                    [v for v in 
+                    self.basic_lineage_tracing_data_simulator.mutation_priors_per_character[i]
+                    .values()]
+                ),
+            )
 
     def test_setup_errors(self):
```
mattjones315, line 168: I'd like to see more tests around passing in mutation rates incorrectly (i.e. a list that's too short/not long enough).

mattjones315 (changes requested):
Looking good! Just a few changes that I think we should implement before merging it in.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
colganwi: Good catch!

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -71,17 +71,29 @@ class Cas9LineageTracingDataSimulator(LineageTracingDataSimulator):
...
+                list of length `number_of_cassettes * size_of_cassette` - each
+                    site and cassettes will mutate at the specified rate.
         state_generating_distribution: Distribution from which to simulate state
             likelihoods. This is only used if mutation priors are not
             specified to the simulator.
         number_of_states: Number of states to simulate
-        mutation_priors: A mapping from state to probability that a user
-            can specify. If this argument is not None, states will not be
-            pulled from the state distribution.
+        mutation_priors: An optional dictionary mapping states to their prior
```
mattjones315, line 86: `mutation_priors` doesn't show up in the argument list. Is this just an instance variable, or do you want it actually to be an argument?  I think you might mean for this to be `state_priors`.

If this is the case, down in the argument list, `state_priors` typing should be updated.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -132,42 +144,70 @@ def __init__(
...
                 )
-            number_of_characters = size_of_cassette * number_of_cassettes
             self.mutation_rate_per_character = [
                 mutation_rate
             ] * number_of_characters
-        else:
-            if len(mutation_rate) != (
-                self.number_of_cassettes * self.size_of_cassette
-            ):
+        elif isinstance(mutation_rate, list):
```
mattjones315, line 157: might be cleaner just to cast `mutation_rate` to `list` rather than throw an error if `mutation_rate` is a numpy array? what do you think?

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
```diff
@@ -132,42 +144,70 @@ def __init__(
...
-                raise DataSimulatorError("Mutation priors do not sum to 1.")
-            self.number_of_states = len(state_priors)
+            if isinstance(state_priors, dict):
+                self.mutation_priors_per_character = [
+                    state_priors
+                ] * number_of_characters
+            else:
+                if len(state_priors) == number_of_characters:
+                    self.mutation_priors_per_character = state_priors
+                elif len(state_priors) ==  self.size_of_cassette:
```
mattjones315, line 193: can we have the same problem above if `state_prior` is a numpy array?

mattjones315 (changes requested):
Looking great! Some small changes requested around the docstring, argument list, and handling of `state_priors` specified as a `np.array`.

cassiopeia/simulator/Cas9LineageTracingDataSimulator.py
colganwi: Yes. Good catch!

mattjones315 (approved):
Looks great! Let's merge it in.

PR #217, opened 2023-08-23
Reorder children in tree

colganwi:
Added `reorder_children` method to `CassiopeiaTree`. This is useful for reordering clades prior to plotting a tree.


mattjones315 (approved):
LGTM!

Happy to merge it in as is, but I'm curious if you think it would be worthwhile to include functionality to optimize leaf ordering with respect to a piece of meta data (e.g., cell type annotation)? I suppose it  could be a separate function that invokes the `reorder_children` method that you implemented.

PR #220, opened 2023-08-28
Clean fix dtype checking in local plotting 

colganwi:


mattjones315 (changes requested):
Looks great - let's add a test though to make sure that both categorical and strings are supported, and otherwise this fails.

mattjones315 (approved):
looks great! thanks for adding these tests, and thanks for handling the matplotlib colormap warning too.

PR #226, opened 2023-09-26
Ambiguous greedy & parallel dissimilarity computation

mattjones315:
This PR implements two important changes.

## Supporting ambiguous alleles in Cassiopeia Greedy algorithm.

Specific changes:
* Enabled greedy splits to be found when states are ambiguous (by flattening ambiguous states)
* Enabled missing data assignment in greedy splits when ambiguous states are present
* A few new utilities around handling ambiguous states, implemented in `cassiopeia.mixins.utilities`
* A few miscellaneous changes, not related to the greedy algorithm, especially around running tests when software is not installed.

## Supporting parallel dissimilarity matrix
We implement a parallel dissimilarity matrix computation. Due to compatibility issues with `numba`, I introduce a wrapper function around the main bones of the dissimilarity map computation, and allow this to operate on batches. I notice a slight slow down for computations that would be numba jit-compatible (on the order of seconds) but I find drastic runtime improvements for computations that are _not_ jit-compatible. This becomes particularly important for cases dealing with ambiguous alleles as currently the cluster dissimilarity function is not able to be compiled in `nopython` mode. Thus, dramatic speedups - roughly proportional the number of threads, ~10x speedup with 10 threads (as one would expect).

I also find that implementing more prescriptive cluster dissimilarity functions (e.g., specific function for `linkage=np.min` and `dissimilarity_function=weighted_hamming_distance`) allows the function to be compiled with `nopython=False, forceobj=True`, which does speed up the computation noticeably. I retain the original cluster computation to keep it backwards compatible and to allow users to experiment with various linkages and dissimilarity functions on cases where performance is not such an issue.


cassiopeia/solver/DistanceSolver.py
```diff
@@ -59,13 +59,15 @@ class DistanceSolver(CassiopeiaSolver.CassiopeiaSolver):
                 "inverse": Transforms each probability p by taking 1/p
                 "square_root_inverse": Transforms each probability by the
                     the square root of 1/p
+        threads: Number of threads to use for dissimilarity map computation.
```
colganwi, line 62: Maybe make this a more general threads argument? CCphylo supports multithreading with the -t argument so we could also set the number of threads used by the tree solver. Unless we anticipate cases where we want a different number of threads used by the dissimilarly computation and tree solver?

cassiopeia/solver/GreedySolver.py
```diff
@@ -236,21 +266,13 @@ def __add_duplicates_to_tree(
         Returns:
             The tree with duplicates added
         """
+        
+        duplicate_mappings = find_duplicate_groups(character_matrix)
```
colganwi, line 270: Nice cleanup!

cassiopeia/solver/NeighborJoiningSolver.py
```diff
@@ -53,13 +53,15 @@ class NeighborJoiningSolver(DistanceSolver.DistanceSolver):
                 Neighbor-Joining algorithm described by Clausen (2023). 
                 Solution is not guaranteed to be exact.
             "ccphylo_nj": CCPhylo implementation of the Neighbor-Joining.
+        threads: Number of threads to use for dissimilarity map computation.
```
colganwi, line 56: See comment above re threads

cassiopeia/solver/dissimilarity_functions.py
```diff
@@ -403,3 +397,100 @@ def cluster_dissimilarity(
         return 0
 
     return result / num_present if normalize else result
+
+
+def cluster_dissimilarity_weighted_hamming_distance_min_linkage(
```
colganwi, line 402: Nice!

colganwi (changes requested):
Great work @mattjones315! The only major comment I have is about using shared memory for the dissimilarity map computation. With large trees I suspect that copying the character matrix for each thread will be slow and may lead to overflow. Based on some quick research I think this could be fixed using [multiprocessing.shared_memory](https://docs.python.org/3/library/multiprocessing.shared_memory.html#module-multiprocessing.shared_memory), but I admit I haven't thought about it too deeply. If you think this is too much work and out of scope I'm happy to table for now.

I also left a few minor comments re formatting and the threads parameter. Happy to clarify anything if needed. Once these are addressed and it the tests are fixed we can merge.

cassiopeia/config.ini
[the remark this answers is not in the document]
mattjones315: Hmm, that is weird. I agree, this should be removed form the commit. I'll look into this.

cassiopeia/data/CassiopeiaTree.py
```diff
@@ -1401,7 +1401,11 @@ def get_mutations_along_edge(
 
         mutations = []
         for i in range(self.n_character):
-            if parent_states[i] != child_states[i]:
+
```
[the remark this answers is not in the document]
mattjones315: It is possible for a parent state to be ambiguous if _all_ of its children have the same ambiguous state. 

But you are right -- if a parent has an ambiguous state, the child must have the same ambiguous state. If a child does _not_ have an ambiguous state, that reflects on the true state of the parent (either 0, if the child's state does not appear in the other child's ambiguous states, or it will resolve the ambiguous state to a particular state `s`).

cassiopeia/data/utilities.py
```diff
@@ -170,12 +207,84 @@ def compute_dissimilarity_map(
...
 
+    if threads > 1:
+        dm = np.zeros(C * (C - 1) // 2, dtype=np.float64)
+        k, m = divmod(len(dm), threads)
+        batches = [
+            np.arange(len(dm))[i * k + min(i, m) : (i + 1) * k + min(i + 1, m)]
+            for i in range(threads)
+        ]
+
+        with multiprocessing.Pool(processes=threads) as pool:
```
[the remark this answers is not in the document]
mattjones315: Great catch and call out. I have to admit, this looks a bit complicated and perhaps out of scope for my current implementation (I've tried this on trees > 10k leaves without too much trouble). I think the main memory bottleneck will be on the NxN dissimilarity matrix, not necessarily on allocating `num_threads` character matrices. 

But I believe I might have a solution, just by subsetting the character matrix to the desired indices for each iteration. I will try implementing this.

cassiopeia/data/utilities.py
```diff
@@ -170,12 +207,84 @@ def compute_dissimilarity_map(
...
+                        )
+                        for batch in batches
+                    ],
+                ),
+            )
+
+        for batch_indices, batch_results in results:
+            dm[batch_indices] = batch_results
+    else:
+        _, dm = __compute_dissimilarity_map_wrapper(dissimilarity_func, cm, np.arange(C*(C-1) // 2), weights, missing_state_indicator, numbaize, ambiguous_present)
```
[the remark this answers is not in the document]
mattjones315: thanks!

cassiopeia/data/utilities.py
```diff
@@ -170,12 +207,84 @@ def compute_dissimilarity_map(
...
+
+    return dm
+
+
+def __compute_dissimilarity_map_wrapper(
```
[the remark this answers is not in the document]
mattjones315: I don't really see how this will be used when invoking from the main `compute_dissimilarity_map` function, and currently we don't have a need for it so I wonder if it will be better to abide by YAGNI ("You Aren't Gonna Need It"), that is we will implement this when we have a use case.

cassiopeia/solver/DistanceSolver.py
mattjones315: Happy to change the docstring here.

cassiopeia/solver/GreedySolver.py
mattjones315: Thank you!

cassiopeia/solver/dissimilarity_functions.py
mattjones315: Thanks you!

cassiopeia/solver/NeighborJoiningSolver.py
mattjones315: Docstring updated.

mattjones315:
Thanks @colganwi for a great review! I've made several of your requested changes, which were very insightful, and I believe I'm now ready for a second review.

One small comment is on the `config.ini` issue in `.gitignore` -- I found that if you specify it in the `.gitignore`, it doesn't get packaged. It's quite a tricky problem. So I removed it from tracking, added a dummy version to `./data` and readded a `cassiopeia/config.ini` to my own personal distribution. Let me know if you have a better idea.

cassiopeia/config.ini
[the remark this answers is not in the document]
colganwi: That solution sounds good to me

cassiopeia/data/utilities.py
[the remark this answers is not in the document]
colganwi: Looks good! I agree the NxN matrix is the bigger problem, but this should offer some benefit with minimal overhead. I don't think there is any good way to reduced `dm` memory. Let's just hope that `dm[batch_indices] = batch_result` is relatively efficient. If this becomes an issue in the future we can deal with it.

cassiopeia/solver/DistanceSolver.py
```diff
@@ -318,8 +322,11 @@ def _setup_ccphylo(self) -> None:
 
         # get ccphylo path
         config = configparser.ConfigParser()
-        config.read(os.path.join(os.path.dirname(__file__),"..","config.ini"))
-        self._ccphylo_path = config.get("Paths","ccphylo_path")
+        config_file = os.path.join(os.path.dirname(__file__),"..","config.ini")
```
colganwi, line 325: change since config.ini has moved

test/solver_tests/ccphylo_solver_test.py
```diff
@@ -48,12 +48,12 @@ def delta_fn(
...
 # only run test if ccphylo_path is specified in config.ini
 config = configparser.ConfigParser()
-config.read(os.path.join(os.path.dirname(__file__),
-                         "..","..","cassiopeia","config.ini"))
-CCPHYLO_CONFIGURED = (config.get("Paths","ccphylo_path") != 
-                      "/path/to/ccphylo/ccphylo")
-print(CCPHYLO_CONFIGURED)
-
+config_path = os.path.join(os.path.dirname(__file__),
+                         "..","..","cassiopeia","config.ini")
```
colganwi, line 52: config.ini has moved

colganwi (approved):
Looks good! Maybe just add some tests for `cluster_dissimilarity_weighted_hamming_distance_min_linkage` and add a note about config.ini to the README

cassiopeia/data/utilities.py
```diff
@@ -136,30 +153,53 @@ def _to_newick_str(g, node):
...
     missing_state_indicator: int = -1,
+    threads: int = 1,
 ) -> np.array:
     """Compute the dissimilarity between all samples
 
     An optimized function for computing pairwise dissimilarities between
     samples in a character matrix according to the dissimilarity function.
 
+
+    TODO: Implement parallelization with a shared memory buffer.
```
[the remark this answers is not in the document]
mattjones315: Thanks for catching!

cassiopeia/data/CassiopeiaTree.py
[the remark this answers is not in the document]
mattjones315: Done.

PR #228, opened 2023-10-18
Leaf Subsampler Speedup

colganwi:
This PR makes the `UniformLeafSubsampler` ~50X faster for large trees by improving `remove_leaves_and_prune_lineages`. Specifically, instead of checking internal nodes many times, we are now using a heap to check internal nodes in order from the bottom of the tree to the top. Additionally, the `__network` now is modified directly which avoids repeated calls to `__check_network_initialized()`.

### DIFF
--- a/cassiopeia/data/CassiopeiaTree.py
+++ b/cassiopeia/data/CassiopeiaTree.py
@@ -19,6 +19,7 @@ import warnings
 from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Union
 
 import ete3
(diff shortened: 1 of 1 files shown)


colganwi:
Also, speeds up the `BirthDeathFitnessSimulator` since a large proportion of this simulations time is spent cleaning up dead lineages.

mattjones315 (approved):
Looks great! Fantastic improvement in runtime and scaling up a lot of our simulation engines.

PR #232, opened 2023-10-30
Newick string branch length support

colganwi:
This PR enables instantiation of the `CassiopeiaTree` with newick strings that contain branch lengths. Previously branch lengths were ignored with all branch lengths being set to 1.


mattjones315 (approved):
LGTM. Just one small comment around adding a more descriptive docstring to the newick->CassiopeiaTree function.

PR #233, opened 2023-11-16
Single thread hybrid solver

colganwi:
Currently, `cas.solver.HybridSolver` cannot be run in parallel because daemonic processes cannot spawn new processes. This PR solves this by modifying `cas.solver.HybridSolver` to not use multiprocessing when `threads = 1`.

### DIFF
--- a/cassiopeia/solver/ILPSolver.py
+++ b/cassiopeia/solver/ILPSolver.py
@@ -126,10 +126,14 @@ class ILPSolver(CassiopeiaSolver.CassiopeiaSolver):
                 " analysis."
             )
 
-        # setup logfile config
-        handler = logging.FileHandler(logfile)
-        handler.setLevel(logging.INFO)
-        logger.addHandler(handler)
+        # configure logger
+        if logfile is not None:
+            file_handler = logging.FileHandler(logfile)
+            file_handler.setLevel(logging.INFO)
+            logger.addHandler(file_handler)
+        logger.ch.setLevel(logging.getLogger().level)
+    
+        # add to logger
         logger.info("Solving tree with the following parameters.")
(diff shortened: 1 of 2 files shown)


cassiopeia/solver/ILPSolver.py
```diff
@@ -272,7 +276,8 @@ def solve(
             cassiopeia_tree.collapse_mutationless_edges(
                 infer_ancestral_characters=True
             )
-        logger.removeHandler(handler)
+
+        #logger.removeHandler(handler)
```
mattjones315, line 280: why do we have this commented out?

mattjones315 (approved):
Looks generally fine to me here, though one comment about a commented-out line. Let's try to not have things like this in the codebase.

Also, can you be a bit more clear about the failure case here? When exactly are you seeing that hybrid cannot be run with multiple threads?

cassiopeia/solver/ILPSolver.py
colganwi: Good catch! I had that commented out since I changed `handler` to `file_handler`

colganwi:
This is for the case of running the hybrid solver in parallel with multiprocessing. Since multiprocessing creates demonic threads these threads cannot then call multiprocessing again to create more threads.

The logging changes enable better control of the logging behavior since the logging output become quite long when hybrid solver is called in parallel. Specifically, `logfile` can be set to None to disable logging and the default `stdout` logging level has been changed to `WARNING`. If you still want the hybrid solver to print to `stdout` you can override the logging level with:

`logging.basicConfig(level=logging.INFO)`

mattjones315:
Hi @colganwi, 

Great, that makes a lot more sense now. I've seen the same issue pop up, so thanks for the PR. Looks great to me and will merge it in.

PR #238, opened 2024-03-07
Lca reconstruct

mattjones315:
Found a small bug in which the reconstructed LCA of a set of ambiguous states would default to 0, even if there was non-zero overlap between the states. This small PR fixes this bug.


colganwi (approved):
Looks good to me!

PR #251, opened 2024-10-24
Added python 3.10 and 3.11 support

colganwi:


colganwi:
@mattjones315 could you take a look at this when you get a chance?

mattjones315 (approved):
Looks good to me - just going to make a few nit-picky changes around import ordering and then we can merge it in. Thanks @colganwi !

PR #255, opened 2025-02-11
Add option to record internal node names when writing tree to newick.

StephenStaklinski:


mattjones315:
Hi @StephenStaklinski,

Thanks very much for using Cassiopeia and for requesting this PR. It sounds like this will be useful if you want to record all the internal nodes, instead of just the leaf names. This changes to `data.utilities.to_newick` look good, but you'll also want to add an option to `CassiopeiaTree.get_newick` to make sure this option is exposed to the user from the CassiopeiaTree API. 

Thanks very much again and please let me know if you need any pointers on how to make this last small change.

Best,
Matt

StephenStaklinski:
Hi @mattjones315, thanks for pointing this out. Those changes should be made now. Let me know if I missed anything else.

mattjones315:
Thanks @StephenStaklinski -- two more slight changes.

First, can you add a docstring update for the function you changed in `CassiopeiaTree`?

Second, can you make sure that all lines in the files you change are at most 80 characters long? You can do this by running `black ./cassioepia/data/CassiopeiaTree.py --line-length 80` for example.

The tests will fail in the Github workflow because we need to update the workflow versions, but they're passing on my side so we'll be good to merge after these two small changes are made.

StephenStaklinski:
Hi @mattjones315, I made those changes. Let me know if anything else should be done. Thanks.

mattjones315 (approved):

PR #262, opened 2025-07-22
Spatial tutorial

mattjones315:
This PR implements the framework for a spatial library. Already, there were pieces of this library such as the `ClonalSpatialDataSimulator`; here, we implement a separate module - `spatial` - which currently implements a spatial imputation procedure.

We also include a tutorial notebook for the spatial data.


cassiopeia/spatial/spatial_utilities.py
```diff
@@ -0,0 +1,116 @@
...
+from typing import Optional, Tuple
+
+import anndata
+import networkx as nx
+import numpy as np
+import pandas as pd
+import squidpy as sq
+
+
+def get_spatial_graph_from_anndata(
```
colganwi, line 14: What value does this function provide? It's just a wrapper around `sq.gr.spatial_neighbors` with a call to `nx.from_numpy_array`. I suggest removing which will remove the squidpy requirement. Users who want to run `impute_spatial_data` can call `sq.gr.spatial_neighbors` first outside of cassiopeia.

cassiopeia/spatial/spatial_imputation.py
```diff
@@ -0,0 +1,125 @@
...
+    if (not spatial_graph) and (not adata):
+        raise Exception(
+            "One of the following must be specified: "
+            "`spatial_graph` or `adata`."
+        )
+
+    if not spatial_graph:
+
+        # create spatial graph if needed
+        spatial_graph = spatial_utilities.get_spatial_graph_from_anndata(
```
colganwi, line 74: I suggest requiring that the spatial_graph is precomputed to eliminate the squidpy dependency

cassiopeia/spatial/spatial_imputation.py
```diff
@@ -0,0 +1,125 @@
...
+            adata,
+            neighborhood_radius=neighborhood_radius,
+            neighborhood_size=neighborhood_size,
+        )
+
+        node_map = dict(
+            zip(
+                range(adata.obsp["spatial_connectivities"].shape[0]),
```
colganwi, line 82: Add `connect_key` argument to allow user to specify this

cassiopeia/spatial/spatial_imputation.py
```diff
@@ -0,0 +1,125 @@
...
+    num_imputation_iterations: int = 1,
+    max_neighbor_distance: float = np.inf,
+    coordinates: Optional[pd.DataFrame] = None,
+) -> pd.DataFrame:
+    """Imputes data based on spatial location.
+
+    This procedure iterates over spots in a given anndata and imputes missing
+    data based on spatial neigborhoods. This procedure is iterative, and can act
+    as an approximation of a message-passing process to smooth over missing
+    data.
```
colganwi, line 33: Please add a bit more detail

cassiopeia/spatial/spatial_utilities.py
```diff
@@ -0,0 +1,116 @@
...
+    spatial_graph = nx.from_numpy_array(adata.obsp["spatial_connectivities"])
+
+    return spatial_graph
+
+
+def impute_single_state(
+    cell: str,
+    character: int,
+    character_matrix: pd.DataFrame,
+    neighborhood_graph: nx.DiGraph = None,
```
colganwi, line 62: Typing

notebooks/spatial.ipynb
colganwi, line 1: Nice job! I appreciate how thorough this tutorial is. You should make it clear that the issue with each "spot" containing multiple cells is specific to spatial transcriptomics data. 

A few spelling issues:
"ambigous" -> "ambiguous".
"spatail" -> "spatial".

cassiopeia/spatial/spatial_imputation.py
```diff
@@ -0,0 +1,125 @@
...
+import anndata
+import networkx as nx
+import numpy as np
+import pandas as pd
+import tqdm
+
+from cassiopeia.spatial import spatial_utilities
+
+
+def impute_spatial_data(
```
colganwi, line 16: Is there a better name for this function? You are using spatial coordinates to impute lineage data not imputing spatial data. I'm fine we the name if you can't think of anything better.

cassiopeia/spatial/spatial_imputation.py
```diff
@@ -0,0 +1,125 @@
...
+    character_matrix: pd.DataFrame,
+    adata: Optional[anndata.AnnData] = None,
```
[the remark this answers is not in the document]
mattjones315: Great idea- done.

cassiopeia/spatial/spatial_imputation.py
mattjones315: I propose we keep it this way for now until we have the treedata class as a dependency in which case a spatial graph constructor will presumably be implemented.

cassiopeia/spatial/spatial_imputation.py
mattjones315: Agreed - renamed to `impute_alleles_from_spatial_data`.

notebooks/spatial.ipynb
mattjones315: Great, thanks!

cassiopeia/spatial/spatial_utilities.py
mattjones315: A few reasons why I have this function:

* This codeblock used in a few places
* I want to have a spatial graph constructor here in case folks don't want to create their spatial graphs and just use squidpy
* Having this abstracted away will be useful in case we want to re-write it.

It might be something we get rid of in the future, but right now I don't think it hurts to have and just adds functionality for users.

cassiopeia/spatial/spatial_imputation.py
mattjones315: Done, thanks!

PR #273, opened 2025-10-16
triplets_correct with networkx 

gloriayma:


src/cassiopeia/critique/compare.py
```diff
@@ -8,110 +8,285 @@
...
+        G1 = to_nx_tree(copy.deepcopy(raw1))
+        G2 = to_nx_tree(copy.deepcopy(raw2))
+
+    elif isinstance(tree1, nx.DiGraph):
+        # nx.DiGraph triplets correct
+        G1 = to_nx_tree(copy.deepcopy(tree1))
+        G2 = to_nx_tree(copy.deepcopy(tree2))
+
+    else:
+        raise TypeError("Unsupported input type. Expected CassiopeiaTree, str (key into tdata.obst), or nx.DiGraph.")
```
mattjones315: Ditto comment above - if we throw an error in a certain situation, this should be made clear in the docstring.

src/cassiopeia/critique/compare.py
```diff
@@ -8,110 +8,285 @@
...
-    unresolved_triplets_correct = defaultdict(int)
-    resolvable_triplets_correct = defaultdict(int)
-    proportion_unresolvable = defaultdict(int)
+    if type(tree1) is not type(tree2):
+        raise TypeError("tree1 and tree2 must be the same type. ")
+
+    if isinstance(tree1, CassiopeiaTree):
+        G1 = to_nx_tree(copy.deepcopy(tree1))
+        G2 = to_nx_tree(copy.deepcopy(tree2))
+
```
mattjones315: Do we need a `to_nx_tree` function?

For CassiopeiaTree objects, you can get this with `tree.get_tree_topology()`. For `treedata`, the objects in the `obst` slot should be networkx objects, no?

mattjones315 (changes requested):
Great job @gloriayma! Left a few comments that we should resolve.

High level comment is that there are a lot of helper functions that we can reduce and refactorize. Also worthwhile to use pre-existing implementations rather than manually implement things like DFS/BFS. In general we can clean up things this way - writing way less code. Check out William's pycea implementation of get_depth, for example: https://github.com/YosefLab/pycea/blob/main/src/pycea/pp/setup_tree.py

src/cassiopeia/critique/compare.py
```diff
@@ -8,110 +8,285 @@
 from collections import defaultdict
```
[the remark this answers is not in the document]
colganwi: I'll change the max length to 100 on the 3.0.0 branch

src/cassiopeia/critique/compare.py
```diff
@@ -4,114 +4,168 @@
...
-        candidate_nodes = depth_to_nodes[depth]
-        total_triplets = sum([T1.get_attribute(v, "number_of_triplets") for v in candidate_nodes])
+    # collapse unifurcations
+    collapse_unifurcations(G1)
+    collapse_unifurcations(G2)
+
+    # annotate depths and per-node triplet counts (on G1)
+    depth_to_nodes = annotate_tree_depths_nx(G1)
+
+    print(G1.nodes(data=True))
```
colganwi: Remove print statement

src/cassiopeia/critique/compare.py
```diff
@@ -4,114 +4,168 @@
...
             A tuple of four dictionaries storing triplet statistics at each depth:
             all_triplets_correct: Total proportion of triplets correct.
             resolvable_triplets_correct: Proportion correct among resolvable triplets.
             unresolved_triplets_correct: Proportion correct among unresolvable triplets.
-            proportion_resolvable: Proportion of triplets that are resolvable at each depth.
+            proportion_unresolvable: Proportion of triplets that are unresolvable at each depth.
+    """
+    if tree2 is None and (key1 is None or key2 is None):
+        raise ValueError("If tree2 is None, both key1 and key2 must be provided.")
+    t1, _ = _get_digraph(tree1, tree_key=key1)
```
colganwi: Set copy = True since you don't want to modify the trees when you calculate the triplets correct.

src/cassiopeia/critique/compare.py
```diff
@@ -4,114 +4,168 @@
...
         else:
-            unresolved_triplets_correct[depth] /= number_unresolvable_triplets
-
-        proportion_unresolvable[depth] = number_unresolvable_triplets / number_of_trials
+            unresolved_triplets_correct[depth] = unres_sum / num_unres
 
-        if proportion_unresolvable[depth] < 1:
-            resolvable_triplets_correct[depth] /= number_of_trials - number_unresolvable_triplets
+        if proportion_unresolvable[depth] < 1.0:
+            denom = number_of_trials - num_unres
```
colganwi: Can be combined with the line below

src/cassiopeia/critique/critique_utilities.py
```diff
@@ -179,3 +181,205 @@ def sample_triplet_at_depth(
...
+    # depth
+    depths = nx.single_source_shortest_path_length(G, root)
+    depth_to_nodes = defaultdict(list)
+    for node, depth in depths.items():
+        G.nodes[node]["depth"] = depth
+        depth_to_nodes[depth].append(node)
+
+    # leaf sizes
+    leaf_sz = {}
+    for u in reversed(list(nx.topological_sort(G))):
```
colganwi: Recalculating the number of leaves below each internal node is O(n^2). Can you come up with an O(n) way to do this? We can go over it tomorrow.

colganwi (changes requested):
Getting close @gloriayma! Most of my comments are formatting and removing redundant code. 

ChatGPT is excellent at doing this! Try copy and pasting this code and asking for a cleaned up and simplified version. You probably won't want to use its output directly but it will give you some ideas of where to trim.

src/cassiopeia/critique/compare.py
gloriayma: got it, do we do the same thing for robinson-foulds?

src/cassiopeia/critique/critique_utilities.py
```diff
@@ -179,3 +181,205 @@ def sample_triplet_at_depth(
         )
 
     return (str(in_group[0]), str(in_group[1]), str(out_group)), out_group
+
+
+def annotate_tree_depths_nx(G: nx.DiGraph) -> dict[int, list[Any]]:
```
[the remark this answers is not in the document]
gloriayma: annotate_tree_depths is defined already on cassiopeia trees

PR #276, opened 2025-10-23
Small parsimony

gloriayma:
Rewrite of small_parsimony in tools module to work with TreeData

### DIFF
--- a/src/cassiopeia/tools/small_parsimony.py
+++ b/src/cassiopeia/tools/small_parsimony.py
@@ -16,13 +16,12 @@ from cassiopeia.mixins.errors import (
 )
 from cassiopeia.typing import TreeLike
 from cassiopeia.utils import (
+    _get_attribute_treelike,
+    _get_cell_meta,
+    _get_children_treelike,
     _get_digraph,
-    get_attribute_treelike,
-    get_cell_meta,
-    get_children_treelike,
+    _set_attribute_treelike,
     get_root,
-    is_leaf_treelike,
-    set_attribute_treelike,
 )
 
 
@@ -108,7 +107,7 @@ def fitch_hartigan_bottom_up(
             or the meta data is not categorical.
     """
     if meta_df is None:
-        meta_df = get_cell_meta(tree)
+        meta_df = _get_cell_meta(tree)
 
     if key not in meta_df.columns:
         raise CassiopeiaError("Key variable does not exist in the metadata for the tree object.")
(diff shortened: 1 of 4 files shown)


src/cassiopeia/tools/small_parsimony.py
```diff
@@ -43,25 +56,31 @@ def fitch_hartigan(
         label_key: Key to add that stores the maximum-parsimony assignment
             inferred from the Fitch-Hartigan top-down refinement.
         copy: Modify the tree in place or not.
+        meta_df: Optional DataFrame containing cell meta data. Only pass in if using networkx DiGraph.
+        treedata_key: If tree is a TreeData object, specify the key corresponding to the tree to process.
 
     Returns:
-            A new CassiopeiaTree if the copy is set to True, else None.
+            A new CassiopeiaTree/TreeData/nx DiGraph + meta_df if the copy is set to True, else None.
```
colganwi: Does not return meta_df so this should be removed from the docstring

src/cassiopeia/tools/small_parsimony.py
```diff
@@ -71,67 +90,80 @@ def fitch_hartigan_bottom_up(
...
+            _set_attribute_treelike(tree, node, add_key, _get_attribute_treelike(g, node, add_key))
+        return tree if copy else None
+    elif isinstance(tree, nx.DiGraph):
+        return g if copy else None
 
-    return cassiopeia_tree if copy else None
+    return tree if copy else None
 
 
 def fitch_hartigan_top_down(
```
colganwi: Same as bottom_up. Private method that accepts nx.Digraph

src/cassiopeia/tools/small_parsimony.py
```diff
@@ -224,13 +266,13 @@ def score_small_parsimony(
...
 
 def fitch_count(
-    cassiopeia_tree: CassiopeiaTree,
-    meta_item: str,
+    tree: CassiopeiaTree | TreeData,
+    key: str,
     root: str | None = None,
     infer_ancestral_states: bool = True,
     state_key: str = "S1",
     unique_states: list[str] | None = None,
```
colganwi: Add tree_key argument

colganwi (changes requested):
@gloriayma were getting close! I've updated the tests to use pytest and treedata. Please review the failing tests and update your code accordingly. You should be updating the tests as you go to ensure that all added functionality is covered.

The major changes here are converting fitch_hartigan_bottom_up and fitch_hartigan_top_down to private methods that take an nx.DiGraph as input. The fitch_hartigan, score_small_parsimony, and fitch_count functions should all have
```
g, _ = _get_digraph(tree, treedata_key)
```
as the first line. This change will make these functions easier to implement since you won't have to think about the type of the tree variable.

